PR #6, opened 2019-03-11
networkd: no networkd handling of tun IFs

t-lo:
This change prevents networkd interference with virtual/tunnel
devices for IPSEC (vti*), docker's virtual interfaces (docker*),
and Calico's virtual interfaces (cali*).

This fixes issues introduced by networkd interference we observed
with IPSEC interfaces in particular (tunnel virtual IP being
de-configured and tunnel route torn down at interface activation
time). It also aims to prevent future issues with docker and Calico.

Signed-off-by: Thilo Fromm <thilo@kinvolk.io>

### DIFF
--- a/systemd/network/calico.network
+++ b/systemd/network/calico.network
@@ -1,0 +1,7 @@
+# Exclude calico ipip tunnel devices from DHCP by default
+
+[Match]
+Name=cali*
+
+[Link]
+Unmanaged=yes
--- a/systemd/network/cni.network
+++ b/systemd/network/cni.network
@@ -1,0 +1,7 @@
+# Exclude kubernetes CNI devices from DHCP by default
+
+[Match]
+Name=cni*
+
+[Link]
+Unmanaged=yes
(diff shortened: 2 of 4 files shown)


dongsupark:
Great!

But I'm not sure about the `docker.network`, because it will conflict with
[the same file in coreos-overlay](https://github.com/flatcar-linux/coreos-overlay/blob/flatcar-master/app-emulation/docker/files/50-docker.network).

And I think other configs like `flannel.network` have been added to `coreos-overlay` instead of the common `init` repo. 
I'm not sure why. Need to track down its git history.

t-lo:
Oh nice, so for Docker we already have a solution. Will remove the file from my pr tomorrow.

Wrt. overlay vs. init repo, I'm open for coniderations.

t-lo:
I've updated the pull request to exclude `docker.network` and to include `cni.network` which solves an  issue ignored by CoreOS upstream: https://github.com/coreos/bugs/issues/1678

My motivation to have the fix in the `init` repo is because the config file which causes the network interference in the first place, `zz-default.network`, is part of the `init` repo, too.

t-lo:
Any chance of getting this into one of the next releases (I realize I just missed a release train)? I don't think the change will have any negative effects on existing use cases.

PR #7, opened 2019-08-02
flatcar-install: Add "-s" flag to install to smallest disk

pothos:
When provisioning a machine, it makes sense to install
Flatcar to the smallest disk.

The disk is chosen by looking at unmounted block devices with a size
greater than 10GB. Which block devices this search should
be restricted to can be configured with the "-I" flag with
a list of major device numbers to search for, or the "-e" flag
with a list of major device numbers to exclude when searching.

### DIFF
--- a/bin/flatcar-install
+++ b/bin/flatcar-install
@@ -20,6 +20,23 @@ default_board() {
     esac
 }
 
+# Finds an unmounted disk with size greater than 10GB and prints it to stdout,
+# otherwise prints an empty line.
+find_smallest_unused_disk() {
+    lsblk -lnpdb -x SIZE -o NAME,SIZE $INCLUDE_MAJOR $EXCLUDE_MAJOR \
+        | (
+        while IFS= read -r line; do
+            drive=$(echo "$line" | awk '{print $1}')
+            size=$(echo "$line" | awk '{print $2}')
+            mountpoints=$(lsblk -ln -o MOUNTPOINT "$drive")
+            if [[ "$size" -gt 10000000000 ]] && [[ -z "$mountpoints" ]]; then
+                echo "$drive"
+            fi
+        done) \
+        | head -1
+    echo "" # Needed to have some output for $(func) when no disk was found
+}
+
 # Everything we do should be user-access only!
 umask 077
 
@@ -43,12 +60,19 @@ for f in /usr/share/oem/oem-release /etc/oem-release; do
     fi
 done
 
-USAGE="Usage: $0 -d <device> [options]
+USAGE="Usage: $0 {{-d <device>}|-s} [options]
 Options:
     -d DEVICE   Install Flatcar Linux to the given device.
+    -s          EXPERIMENTAL: Install Flatcar Linux to the smallest unmounted disk found
+                (min. size 10GB). It is recommended to use it with -e or -I to filter the
+                block devices by their major numbers. E.g., -e 7 to exclude loop devices
+                or -I 8,259 for certain disk types. Read more about the numbers here:
+                https://www.kernel.org/doc/Documentation/admin-guide/devices.txt.
     -V VERSION  Version to install (e.g. current) [default: ${VERSION_ID}].
     -B BOARD    Flatcar Linux board to use [default: ${BOARD}].
     -C CHANNEL  Release channel to use (e.g. beta) [default: ${CHANNEL_ID}].
+    -I|e <M,..> EXPERIMENTAL (used with -s): List of major device numbers to in-/exclude
+                when finding the smallest disk.
     -o OEM      OEM type to install (e.g. ami) [default: ${OEM_ID:-(none)}].
     -c CLOUD    Insert a cloud-init config to be executed on boot.
     -i IGNITION Insert an Ignition config to be executed on boot.
@@ -239,21 +263,24 @@ IMAGE_FILE=
 KEYFILE=
 VERSION_SUMMARY='Flatcar Linux'
 
-while getopts "V:B:C:d:o:c:i:t:b:k:f:nyvh" OPTION
+while getopts "V:B:C:I:d:o:c:e:i:t:b:k:f:nsyvh" OPTION
 do
     case $OPTION in
         V) VERSION_ID="$OPTARG"; VERSION_SPECIFIED=1 ;;
         B) BOARD="$OPTARG" ;;
         C) CHANNEL_ID="$OPTARG"; CHANNEL_SPECIFIED=1 ;;
+        I) INCLUDE_MAJOR="-I $OPTARG" ;;
         d) DEVICE="$OPTARG" ;;
         o) OEM_ID="$OPTARG" ;;
         c) CLOUDINIT="$OPTARG" ;;
+        e) EXCLUDE_MAJOR="-e $OPTARG" ;;
         i) IGNITION="$OPTARG" ;;
         t) ;; # compatibility option; previously set TMPDIR
         b) BASE_URL="${OPTARG%/}" ;;
         k) KEYFILE="$OPTARG" ;;
         f) IMAGE_FILE="$OPTARG" ;;
         n) COPY_NET=1;;
+        s) INSTALL_TO_SMALLEST=1 ;;
         y) DRY_RUN=1 ;;
         v) set -x ;;
         h) echo "$USAGE"; exit;;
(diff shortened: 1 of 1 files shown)


bin/flatcar-install
```diff
@@ -20,6 +20,20 @@ default_board() {
...
+    lsblk -lnpdb -x SIZE -o NAME,SIZE $INCLUDE_MAJOR $EXCLUDE_MAJOR \
+        | (
+        while IFS= read -r line || [ "$line" ]; do
+            size=$(echo "$line" | awk '{print $2}')
+            if [[ "$size" -gt 10000000000 ]]; then
+                echo "$line"
+            fi
+        done) \
+        | head -1 | awk '{print $1}'
+    echo
```
rata: This is needed in case no device is found, right? Can you please add a comment explaining why this echo is needed and is not just "lost"?

bin/flatcar-install
```diff
@@ -20,6 +20,20 @@ default_board() {
     esac
 }
 
+# Find a disk with size greater than 10GB
+find_smallest_disk() {
```
rata: Can you explain the output? Something like: returns the smallest device with more than 10G or "" if none is found?

bin/flatcar-install
rata: Will this also work if it's `echo ""`? Maybe that is more clear that we want to return empty?

bin/flatcar-install
```diff
@@ -276,12 +299,20 @@ Settings:
...
-if [[ -z "${DEVICE}" ]]; then
-    echo "$0: No target block device provided, -d is required." >&2
+if [[ -z "${DEVICE}" ]] && [[ -z "${INSTALL_TO_SMALLEST}" ]]; then
+    echo "$0: No target block device provided, either -d or -s is required." >&2
     echo "$USAGE" >&2
     exit 1
 fi
 
+if [[ -n "${INSTALL_TO_SMALLEST}" ]]; then
+    DEVICE="$(find_smallest_disk)"
```
rata: What about having defaults for majors to include if none is set? Would it make sense?

Like for sure 8,259 and probably some other

bin/flatcar-install
```diff
@@ -43,12 +57,18 @@ for f in /usr/share/oem/oem-release /etc/oem-release; do
...
     -d DEVICE   Install Flatcar Linux to the given device.
+    -s          Install Flatcar Linux to the smallest disk found (min. 10GB).
+                It is recommended to use it with -e or -I to filter the block
+                devices by their major numbers. E.g., -e 7 to exclude loop devices
+                or -I 8,259 for certain disk types. Read more about the numbers here:
+                https://www.kernel.org/doc/Documentation/admin-guide/devices.txt.
     -V VERSION  Version to install (e.g. current) [default: ${VERSION_ID}].
     -B BOARD    Flatcar Linux board to use [default: ${BOARD}].
     -C CHANNEL  Release channel to use (e.g. beta) [default: ${CHANNEL_ID}].
+    -I|e <M,..> List of major device numbers to in-/exclude when finding the smallest disk.
```
rata: While I like the flexibility, I'm not sure we want to keep the functionality as minimum as possible exposed to the user, so it is also trivial to remove this in the future

I doubt if it's better to add this ir not. I'll let other decide :)

bin/flatcar-install
```diff
@@ -43,12 +57,18 @@ for f in /usr/share/oem/oem-release /etc/oem-release; do
     fi
 done
 
-USAGE="Usage: $0 -d <device> [options]
+USAGE="Usage: $0 {{-d <device>}|-s} [options]
 Options:
     -d DEVICE   Install Flatcar Linux to the given device.
+    -s          Install Flatcar Linux to the smallest disk found (min. 10GB).
```
rata: I'd add an "EXPERIMENTAL" or also mention that it might be removed at any time

bin/flatcar-install
pothos: Done.

bin/flatcar-install
pothos: Yes, thanks, I'll add a comment.

pothos: Ok :)

bin/flatcar-install
pothos: The problem is that this is too limiting - major numbers are too many with no category system. If we need to change this we have to rebuild the whole OS. Users should know what disks they use in their servers and just pass it with `-I`. It's easy for them to test it on their side fist with `lsblk -I 1,2,3,…`). Or they know what they want to avoid (e.g., just loop devices) and can simply provide `-e 7` and don't care about future disk type changes.

bin/flatcar-install
pothos: Added the EXPERIMENTAL marker

bin/flatcar-install
rata: Sorry I wasn't clear, I was saying a default when no flag (include or exclude) has been specified. For example on my laptop it list the cdroom (yeah, old laptop :P), not sure such things make sense by default.

But I'm okay, it's experimental and we can change it later if needed :)

bin/flatcar-install
```diff
@@ -20,6 +20,21 @@ default_board() {
     esac
 }
 
+# Finds a disk with size greater than 10GB and prints it to stdout,
+# otherwise prints an empty line.
+find_smallest_disk() {
+    lsblk -lnpdb -x SIZE -o NAME,SIZE $INCLUDE_MAJOR $EXCLUDE_MAJOR \
+        | (
+        while IFS= read -r line || [ "$line" ]; do
```
rata: Not sure I follow why the `||` is needed, is this to handle fine that the script is run with `-e` and read will exit with error when EOF is hit?

Maybe another comment to clear that out? (sorry :D)

bin/flatcar-install
```diff
@@ -43,12 +58,18 @@ for f in /usr/share/oem/oem-release /etc/oem-release; do
...
     -d DEVICE   Install Flatcar Linux to the given device.
+    -s          EXPERIMENTAL: Install Flatcar Linux to the smallest disk found (min. 10GB).
+                It is recommended to use it with -e or -I to filter the block
+                devices by their major numbers. E.g., -e 7 to exclude loop devices
+                or -I 8,259 for certain disk types. Read more about the numbers here:
+                https://www.kernel.org/doc/Documentation/admin-guide/devices.txt.
     -V VERSION  Version to install (e.g. current) [default: ${VERSION_ID}].
     -B BOARD    Flatcar Linux board to use [default: ${BOARD}].
     -C CHANNEL  Release channel to use (e.g. beta) [default: ${CHANNEL_ID}].
+    -I|e <M,..> List of major device numbers to in-/exclude when finding the smallest disk.
```
rata: This shouldn't be experimental too?

rata:
> Just a thought, do we want this feature to be available for everyone? Maybe we should simply keep it on the separated branch just for packet for now? Or at least until they have time to test it?

Not sure what is easier in the current workflow, I'll let @dongsupark decide :)

bin/flatcar-install
pothos: Ah, that's tricky: `while read…` works in general if the output ends with a newline character. But as soon as the final newline character is not there, the last line will be skipped. Because this is can cause issues if the code is reused, I make sure that this case is handled even though we would not hit this right now for `lsblk` but maybe if another filter is added in between.

rata: Oh, then I'd vote to remove it... is simpler and if something more complicated in the future is needed, that hits that case, let's leave that for the ones doing the messy thing :-D

pothos: Ok, removed it.

bin/flatcar-install
pothos: Yes, done

bin/flatcar-install
pothos: We could try to assemble a list of devices to ignore by default and detect if `-I` is used because it's mutually exclusive. But I think the CD drive will not be used anyway because it's size is too small. And I think that the `-s` flag would only be used if the people have a fixed set of drives they are aware of and don't care where it is installed except for choosing the smallest.

dongsupark:
I've not read the code yet.

> Just a thought, do we want this feature to be available for everyone? Maybe we should simply keep it on the separated branch just for packet for now? Or at least until they have time to test it?

Yes, I agree.
Let's keep it on a branch, so that Packet people could use it for now.

pothos:
I've added a check to ignore mounted disks because this would lead to either data loss or a broken installation if the current filesystems write to the same disk (and `wipefs` would anyway refuse to run later).

pothos:
I've changed the merge target to be `install-to-smallest-disk`.

bin/flatcar-install
rata: Okay, let's keep it without defaults then :)

PR #16, opened 2020-02-12
systemd: check both OEM IDs for flatcar and coreos

dongsupark:
For backwards compatibilities, we need to check both OEM IDs in the kernel command line, `flatcar.oem.id` and `coreos.oem.id`.

Partly addresses https://github.com/flatcar-linux/Flatcar/issues/16.


dongsupark:
Updated it so it can simply have 2 different network files for both flatcar and coreos.

pothos (approved):

PR #18, opened 2020-05-04
bin/flatcar-install: Describe that setting OEM uses an OEM image

pothos:
The description for -o OEM talks about an OEM type that is installed.
Add that this means that a different image will be used. This makes
it easier for users to relate this to the images we offer on the
release server.

# How to use/test

`bin/flatcar-install -h`

Note: Pick for all channels.


sayanchowdhury (approved):
LGTM

margamanterola (approved):
LGTM

PR #23, opened 2020-07-20
Add a Match section for systemd to not ignore

sayanchowdhury:
# Add a Match section for systemd to not ignore

systemd 245 ignores .link and .network files if the Match section is empty. More details in the [NEWS](https://github.com/systemd/systemd/blob/master/NEWS#L830-L833) entry. 

This should be merged with https://github.com/flatcar-linux/coreos-overlay/pull/448

Signed-off-by: Sayan Chowdhury <sayan.chowdhury2012@gmail.com>

# How to use

```bash
emerge-amd54-usr coreos-base/coreos-init
```

# Testing done

Without these options, the SSH to the machine was broken, and no packets were received by the machine. This fixes the issue.

### DIFF
--- a/systemd/network/yy-pxe.network
+++ b/systemd/network/yy-pxe.network
@@ -1,4 +1,5 @@
 [Match]
+Name=*
 KernelCommandLine=!root
 
 [Network]
--- a/systemd/network/zz-default.network
+++ b/systemd/network/zz-default.network
@@ -1,6 +1,9 @@
 [Network]
 DHCP=yes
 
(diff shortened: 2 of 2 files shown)


pothos (commented):
Looks good. Same as in https://github.com/flatcar-linux/bootengine/pull/16 I would also expect `Name=*` to be set in `systemd/network/yy-pxe.network`

pothos (approved):

dongsupark (approved):

PR #24, opened 2020-08-27
flatcar-install: support installation on multipath devices

simonello:
# Allow installation to multipath target

I got the error "Target block device /dev/mapper/mpatha is not a full disk". I had to change line 360 in the /usr/bin/coreos-install script and add mpath. Now I can install on /dev/mapper/mpatha

https://github.com/flatcar-linux/coreos-overlay/pull/528

# How to use

systemctl start multipathd
coreos-install -d /dev/mapper/mpatha

# Testing done

it installs and comes up on a reboot.

### DIFF
--- a/bin/flatcar-install
+++ b/bin/flatcar-install
@@ -357,7 +357,7 @@ if [[ -n "${INSTALL_TO_SMALLEST}" ]]; then
     fi
 fi
 
-if ! [[ $(lsblk -n -d -o TYPE "${DEVICE}") =~ ^(disk|loop|lvm)$ ]]; then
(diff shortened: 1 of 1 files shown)


dongsupark (approved):
Thanks for the PR.
The code looks good. 
We would also need another PR in coreos-overlay, to pull in the new commit in the init repo.
For now I will let other maintainers have a look as well.

pothos (approved):

dongsupark:
Filed a PR https://github.com/flatcar-linux/coreos-overlay/pull/561

PR #25, opened 2020-08-31
flatcar-install: update public key

pothos:
The current subkey will expire soon and a new one was generated.

# How to use

Run `flatcar-install` for a version signed with the new subkey.

# Testing done

None


vbatts:
:+1: 
Future: we need to revisit this workflow, so it's based more on trusting the public key of the root key, than the expiration of the subkeys. But for now at least we're rotating this before it expires. And this only would affect the install script key, not the update-signing key.

t-lo (approved):
LGTM, thank you!

PR #27, opened 2020-09-21
scripts/motdgen: Check LTS configuration and warn on login

pothos:
# How to use

# Testing done

Copied the file to a running instance and varied all combinations to check the logic.
```
Flatcar Container Linux by Kinvolk lts (2512.2.1)
/etc/flatcar/update.conf: GROUP=lts on the public update server can't be used directly, use GROUP=lts-STREAM or your own update server with a managed lts group.
core@localhost ~ $ 
```

### DIFF
--- a/scripts/motdgen
+++ b/scripts/motdgen
@@ -17,3 +17,11 @@ echo -e "\e[${ANSI_COLOR}m${NAME}\e[39m ${GROUP} (${VERSION})" > /run/flatcar/mo
 if [[ -d "/etc/motd.d" ]]; then
 	cat /etc/motd.d/*.conf 2>/dev/null >> /run/flatcar/motd || true
 fi
+
+if [[ "${GROUP}" =~ "lts" ]] && { [[ "${DOWNLOAD_USER}" = "" ]] || [[ "${DOWNLOAD_PASSWORD}" = "" ]] ; }; then
+	echo "/etc/flatcar/update.conf: DOWNLOAD_USER and DOWNLOAD_PASSWORD are missing to download LTS updates." >> /run/flatcar/motd
+fi
+if [[ "${GROUP}" = "lts" ]] && [[ "${SERVER}" = "https://public.update.flatcar-linux.net/v1/update/" ]]; then
+	echo -n "/etc/flatcar/update.conf: GROUP=lts on the public update server is a placeholder and can't be used directly, " >> /run/flatcar/motd
+	echo "use GROUP=lts-STREAM or your own update server with a managed 'lts' group." >> /run/flatcar/motd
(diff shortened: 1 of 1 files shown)


scripts/motdgen
```diff
@@ -17,3 +17,11 @@ echo -e "\e[${ANSI_COLOR}m${NAME}\e[39m ${GROUP} (${VERSION})" > /run/flatcar/mo
 if [[ -d "/etc/motd.d" ]]; then
 	cat /etc/motd.d/*.conf 2>/dev/null >> /run/flatcar/motd || true
 fi
+
+if [[ "${GROUP}" =~ "lts" ]] && { [[ "${DOWNLOAD_USER}" = "" ]] || [[ "${DOWNLOAD_PASSWORD}" = "" ]] ; }; then
+	echo "/etc/flatcar/update.conf: DOWNLOAD_USER and DOWNLOAD_PASSWORD are missing to download LTS updates." >> /run/flatcar/motd
+fi
+if [[ "${GROUP}" = "lts" ]] && [[ "${SERVER}" = "https://public.update.flatcar-linux.net/v1/update/" ]]; then
+	echo -n "/etc/flatcar/update.conf: GROUP=lts on the public update server can't be used directly, " >> /run/flatcar/motd
```
margamanterola: I think this error message is not particularly clear. What do you mean with "can't be used *directly*"?

scripts/motdgen
```diff
@@ -17,3 +17,11 @@ echo -e "\e[${ANSI_COLOR}m${NAME}\e[39m ${GROUP} (${VERSION})" > /run/flatcar/mo
...
+	echo "use GROUP=lts-STREAM or your own update server with a managed lts group." >> /run/flatcar/motd
```
margamanterola: What would STREAM be here?

margamanterola (commented):
There's a typo in the commit message that I fixed for the PR title, but you'll need to fix for the message itself.

scripts/motdgen
pothos: STREAM will be 2020, 2021, etc.

scripts/motdgen
pothos: I added a reason now: _GROUP=lts on the public update server is a placeholder and can't be used directly_

scripts/motdgen
margamanterola: And the documentation will refer to this as STREAM and not YEAR? I just find it a bit confusing, but I guess if the documentations uses the same term, then it's ok.

margamanterola (approved):
LGTM

vbatts:
LGTM

PR #28, opened 2020-10-13
Do not create empty /etc/coreos symlink

pothos:
The symlink had no target and was only working on the final image after
the target folder got created by some other component. Now the target
folder and symlink is created by a tmpfile directive in baselayout and
the symlink here conflicts with the baselayout package's file.
Remove it since it is a duplicate.

# How to use

Goes along with https://github.com/flatcar-linux/baselayout/pull/7

# Testing done

See linked PR.


margamanterola (approved):
LGTM

pothos:
I will include it in all maintenance branches.

PR #34, opened 2021-02-02
scripts/motdgen: Add OEM information to motd output

pothos:
The message-of-the-day login greeter shows basic information about the
OS but so far did not give a hint whether an OEM variant is used and
which (e.g., "Flatcar Container Linux by Kinvolk stable (2605.12.0)").
Append the OEM name to it, like "for Amazon EC2".

# How to use

Build an OEM image and see that the name is shown at login

# Testing done

Only manual execution of the changed script lines

### DIFF
--- a/scripts/motdgen
+++ b/scripts/motdgen
@@ -12,7 +12,11 @@ source /usr/lib/os-release
 
 mkdir -p /run/flatcar
 ln -sf flatcar /run/coreos
-echo -e "\e[${ANSI_COLOR}m${NAME}\e[39m ${GROUP} (${VERSION})" > /run/flatcar/motd
+OEM=""
+if [ -f /usr/share/oem/oem-release ]; then
+  OEM="$(source /usr/share/oem/oem-release; echo " for $NAME")"
+fi
+echo -e "\e[${ANSI_COLOR}m${NAME}\e[39m ${GROUP} (${VERSION})${OEM}" > /run/flatcar/motd
 
 if [[ -d "/etc/motd.d" ]]; then
 	cat /etc/motd.d/*.conf 2>/dev/null >> /run/flatcar/motd || true
--- a/scripts/motdgen
+++ b/scripts/motdgen
@@ -16,7 +16,7 @@ OEM=""
 if [ -f /usr/share/oem/oem-release ]; then
   OEM="$(source /usr/share/oem/oem-release; echo " for $NAME")"
 fi
-echo -e "\e[${ANSI_COLOR}m${NAME}\e[39m ${GROUP} (${VERSION})${OEM}" > /run/flatcar/motd
+echo -e "\e[${ANSI_COLOR}m${NAME}\e[39m ${GROUP} ${VERSION}${OEM}" > /run/flatcar/motd
 
 if [[ -d "/etc/motd.d" ]]; then
 	cat /etc/motd.d/*.conf 2>/dev/null >> /run/flatcar/motd || true


scripts/motdgen
```diff
@@ -12,7 +12,11 @@ source /usr/lib/os-release
 
 mkdir -p /run/flatcar
 ln -sf flatcar /run/coreos
-echo -e "\e[${ANSI_COLOR}m${NAME}\e[39m ${GROUP} (${VERSION})" > /run/flatcar/motd
+OEM="$(grep "^NAME=" /usr/share/oem/oem-release 2> /dev/null | cut -d '"' -f 2)"
```
margamanterola: This works and normally I'm a cut fan, but parsing variables seems not too great... I found this alternative syntax, that also works and doesn't do variable parsing with cut:
```
OEM="$( source /usr/share/oem/oem-release; echo ", $NAME")"
```

But I guess it only works if the file exists... Is that the reason why you have the conditional afterwards, that for the basic image the file doesn't exist?

Maybe:
```
OEM=""
if [[ -f /usr/share/oem/oem-release ]]; then
  OEM="$( source /usr/share/oem/oem-release; echo ", $NAME")"
fi
```

This works in my tests and avoids cutting with the quote which seems fragile. What do you think?

scripts/motdgen
```diff
@@ -12,7 +12,11 @@ source /usr/lib/os-release
...
+if [ "${OEM}" != "" ]; then
+  OEM=", ${OEM}"
+fi
+echo -e "\e[${ANSI_COLOR}m${NAME}\e[39m ${GROUP} (${VERSION})${OEM}" > /run/flatcar/motd
```
margamanterola: The end string is then: 
`Flatcar Container Linux by Kinvolk stable (2605.12.0), Amazon EC2`

It might look better if we use "for": 
`Flatcar Container Linux by Kinvolk stable (2605.12.0) for Amazon EC2`

But it won't fit the name scheme that we want for the Pro images. We want the offer to be "Flatcar Container Linux Pro for ...". And it doesn't fit:
`Flatcar Container Linux by Kinvolk stable (2605.12.0), Amazon EC2 Pro`
`Flatcar Container Linux by Kinvolk stable (2605.12.0)  for Amazon EC2 Pro`
`Flatcar Container Linux by Kinvolk stable (2605.12.0), Pro for AWS`

So... I think we need to put the string in between the "Flatcar Container Linux " and "by Kinvolk". Which we could do with some text parsing but that seems ugly and unnecessary given that we own the original string.  If we want to be "fork friendly" we could do:

```
COMPNAME=""
if [[ NAME = "Flatcar Container Linux by Kinvolk" ]]; then
  NAME="Flatcar Container Linux"
  COMPNAME="by Kinvolk"
fi
```

And then:
```
echo -e "\e[${ANSI_COLOR}m${NAME}${OEM} ${COMPNAME}\e[39m ${GROUP} (${VERSION})" > /run/flatcar/motd
```

With `OEM= Pro for AWS` this would render:
`Flatcar Container Linux Pro for AWS by Kinvolk stable (2605.12.0)`

But while this works for the Pro thing, it doesn't work so well for the non-Pro OEMs. e.g. with `OEM= Equinix Metal`:
`Flatcar Container Linux Equinix Metal by Kinvolk stable (2605.12.0)`

So, all this rambling to say, I don't think the oem-release NAME field is enough :-/. I think we want to have a separate offer name that stores Pro or LTS, so that we can then create "Pro for X", "LTS for X" or "for X".

What do you think?

margamanterola (commented):
Thanks for working on this. Unfortunately, I'm not sure this ends up creating the string we want :(

scripts/motdgen
pothos: Yes, I think it's nicer to have a separate field for Pro, also the `for X` looks good.

I'm not sure about splitting up the first part of it… how about that:
`Flatcar Container Linux by Kinvolk (Pro) stable 2605.12.0 for Amazon EC2`

scripts/motdgen
pothos: Good trick with the subshell to not overwrite the existing `NAME`. This was the reason why I resorted to grep.

scripts/motdgen
margamanterola: Yeah, I guess this is near enough to what we want. But we'll need to add a new field for the offer type ("", "Pro" or "LTS")...

Given that we mentioned we'll have separate /usr partitions for Pro, we might want to actually make the PRETTY_NAME be this, and drop the codename that we don't use?

margamanterola: Oh well, I guess `Flatcar Container Linux by Kinvolk stable 2605.12.0 for Amazon EC2 (Pro) ` is good enough for now and we can do the other thing after we have a separate /usr partition.

margamanterola (approved):
LGTM

scripts/motdgen
pothos: We can ignore LTS here because it is a channel, not a variant of another channel

pothos: Removed the brackets around the version, the rest can be done in other PRs.

PR #35, opened 2021-02-12
bin/flatcar-install: add parameters to make wget more resilient

pothos:
When the network has temporary issues it can happen that wget fails to
establish a connection or complete the transfer, or worse, get stuck
in a TCP connection which is not moving forward.
Add timeout and retry parameters to prevent unnecessary failure if a
retry fixes the problem and also to prevent the connection from getting
stuck.

# How to use

Use `flatcar-install -s` from a PXE-booted Flatcar image to install to disk, check that it works and is even robust against the network dropping packets for the specified timeout of 20 seconds.

# Testing done

Ran the `wget` command line on a test file while breaking my local connection. Also tested that `sudo flatcar-install -s` still works as expected.

### DIFF
--- a/bin/flatcar-install
+++ b/bin/flatcar-install
@@ -339,6 +339,8 @@ IMAGE_FILE=
 KEYFILE=
 VERSION_SUMMARY='Flatcar Container Linux'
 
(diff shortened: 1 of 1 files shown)


margamanterola (approved):
LGTM

PR #37, opened 2021-03-10
Extend the Flatcar Container Linux installer script to include a -D option for local image download

### LINKED ISSUE #248: add flag to flatcar-install to only download the image file
**Current situation**

I want to be able to package flatcar-install and a specific version of the image (e.g. https://stable.release.flatcar-linux.net/amd64-usr/current/flatcar_production_image.bin.bz2) in a docker container so it can be easily installed by a [tinkerbell workflow action](https://github.com/tinkerbell/tink) in a air-gap environment.

**Impact**

[flatcar-master/bin/flatcar-install](https://github.com/flatcar-linux/init/blob/flatcar-master/bin/flatcar-install) need to be changed in a backwards compatible way.

**Ideal future situation**

add a switch to flatcar-install that only downloads the image and exits. for example with:

```bash
flatcar-install -D image.bz2
```

Or do not even have an argument, just download the file to the current directory, e.g.:

```bash
flatcar-install -D
```

And a script of mine can look for the actual filename and later pass it to `flatcar-install -f`.

Then later I can install it with:

```bash
flatcar-install -f image.gz2 -d /dev/sda
```

Having this flag is easier than implementing the whole download/verification logic myself; I can just use something like `ADD https://github.com/flatcar-linux/init/blob/build-1688.4.0/bin/flatcar-install` in my Dockerfile. What do you think?

jasonbraganza:
# Title: Added flag to flatcar-install to only download the image file, as required by  issue #248

The `install_from_url` script was refactored.   
We now have 3 functions  
- `prep_url` handles setting of variables and gpg keys for the download  
- `install_from_url` does the original task of downloading and writing to a disk    
- `download_from_url` downloads the bz2 image to the current folder and saves it.  

Closes https://github.com/kinvolk/Flatcar/issues/248

# How to use

run `flatcar-install -D` and the script should download the image to the current folder

# Testing done

Have tried downloading the image across two computers.
Has successfully executed on one.  
Still downloading on the other (results expected)

### DIFF
--- a/bin/flatcar-install
+++ b/bin/flatcar-install
@@ -67,6 +67,7 @@ done
 USAGE="Usage: $0 {{-d <device>}|-s} [options]
 Options:
     -d DEVICE   Install Flatcar Container Linux to the given device.
+    -D          Download a Flatcar Container Linux image to the current directory, without installing it
     -s          EXPERIMENTAL: Install Flatcar Container Linux to the smallest unmounted disk found
                 (min. size 10GB). It is recommended to use it with -e or -I to filter the
                 block devices by their major numbers. E.g., -e 7 to exclude loop devices
@@ -341,12 +342,13 @@ VERSION_SUMMARY='Flatcar Container Linux'
 
 WGET_ARGS="--tries 10 --timeout=20 --retry-connrefused"
 
-while getopts "V:B:C:I:d:o:c:e:i:t:b:k:f:nsyvh" OPTION
+while getopts "V:B:C:DI:d:o:c:e:i:t:b:k:f:nsyvh" OPTION
 do
     case $OPTION in
         V) VERSION_ID="$OPTARG"; VERSION_SPECIFIED=1 ;;
         B) BOARD="$OPTARG" ;;
         C) CHANNEL_ID="$OPTARG"; CHANNEL_SPECIFIED=1 ;;
+        D) DOWNLOAD_ONLY='true';;
         I) INCLUDE_MAJOR="-I $OPTARG" ;;
         d) DEVICE="$OPTARG" ;;
         o) OEM_ID="$OPTARG" ;;
@@ -380,51 +382,55 @@ Settings:
     IMAGE:    ${IMAGE_FILE:-(none)}"
 }
 
-# Device is required, must not be a partition, must be writable
-if [[ -z "${DEVICE}" ]] && [[ -z "${INSTALL_TO_SMALLEST}" ]]; then
-    echo "$0: No target block device provided, either -d or -s is required." >&2
-    echo "$USAGE" >&2
-    exit 1
-fi
+# Run Device checks, ignore if the download only flag is present
 
-if [[ -n "${INSTALL_TO_SMALLEST}" ]]; then
-    DEVICE="$(find_smallest_unused_disk)"
-    if [[ -z "${DEVICE}" ]]; then
-        echo "$0: Could not find smallest disk to install (min. 10GB)." >&2
-        exit 1
-    fi
-fi
-
-if ! [[ $(lsblk -n -d -o TYPE "${DEVICE}") =~ ^(disk|mpath|loop|lvm)$ ]]; then
-    echo "$0: Target block device (${DEVICE}) is not a full disk." >&2
-    exit 1
-fi
-
-if [[ ! -w "${DEVICE}" ]]; then
-    echo "$0: Target block device (${DEVICE}) is not writable (are you root?)" >&2
-    exit 1
-fi
-
-if [[ -n "${CLOUDINIT}" ]]; then
-    if [[ ! -f "${CLOUDINIT}" ]]; then
-        echo "$0: Cloud config file (${CLOUDINIT}) does not exist." >&2
+if [[ -z "${DOWNLOAD_ONLY}" ]]; then
+    # Device is required, must not be a partition, must be writable
+    if [[ -z "${DEVICE}" ]] && [[ -z "${INSTALL_TO_SMALLEST}" ]]; then
+        echo "$0: No target block device provided, either -d or -s is required. Or use -D to download image." >&2
+        echo "$USAGE" >&2
         exit 1
     fi
 
-    if type -P coreos-cloudinit >/dev/null; then
-        if ! coreos-cloudinit -from-file="${CLOUDINIT}" -validate; then
-            echo "$0: Cloud config file (${CLOUDINIT}) is not valid." >&2
+    if [[ -n "${INSTALL_TO_SMALLEST}" ]]; then
+        DEVICE="$(find_smallest_unused_disk)"
+        if [[ -z "${DEVICE}" ]]; then
+            echo "$0: Could not find smallest disk to install (min. 10GB)." >&2
             exit 1
         fi
-    else
-        echo "$0: coreos-cloudinit not found. Could not validate config. Continuing..." >&2
     fi
-fi
 
-if [[ -n "${IGNITION}" ]]; then
-    if [[ ! -f "${IGNITION}" ]]; then
-        echo "$0: Ignition config file (${IGNITION}) does not exist." >&2
+    if ! [[ $(lsblk -n -d -o TYPE "${DEVICE}") =~ ^(disk|mpath|loop|lvm)$ ]]; then
+        echo "$0: Target block device (${DEVICE}) is not a full disk." >&2
         exit 1
+    fi
+
+    if [[ ! -w "${DEVICE}" ]]; then
+        echo "$0: Target block device (${DEVICE}) is not writable (are you root?)" >&2
+        exit 1
+    fi
+
+    if [[ -n "${CLOUDINIT}" ]]; then
+        if [[ ! -f "${CLOUDINIT}" ]]; then
+            echo "$0: Cloud config file (${CLOUDINIT}) does not exist." >&2
+            exit 1
+        fi
+
+        if type -P coreos-cloudinit >/dev/null; then
+            if ! coreos-cloudinit -from-file="${CLOUDINIT}" -validate; then
+                echo "$0: Cloud config file (${CLOUDINIT}) is not valid." >&2
+                exit 1
+            fi
+        else
+            echo "$0: coreos-cloudinit not found. Could not validate config. Continuing..." >&2
+        fi
+    fi
+
+    if [[ -n "${IGNITION}" ]]; then
+        if [[ ! -f "${IGNITION}" ]]; then
+            echo "$0: Ignition config file (${IGNITION}) does not exist." >&2
+            exit 1
+        fi
     fi
 fi
 
@@ -480,12 +486,12 @@ function install_from_file() {
     VERSION_SUMMARY+=" (from ${IMAGE_FILE})"
 }
 
-function install_from_url() {
+function prep_url(){
     # Ensure that required executables exist before proceeding
     type -P wget >/dev/null || { echo 'Missing wget!' >&2 ; exit 1 ; }
     type -P gpg >/dev/null || { echo 'Missing gpg!' >&2 ; exit 1 ; }
 
-    local IMAGE_NAME="flatcar_production_image.bin.bz2"
+    IMAGE_NAME="flatcar_production_image.bin.bz2"
     if [[ -n "${OEM_ID}" ]]; then
         IMAGE_NAME="flatcar_production_${OEM_ID}_image.bin.bz2"
     fi
@@ -515,10 +521,9 @@ function install_from_url() {
         echo "Current version of Flatcar Container Linux ${CHANNEL_ID} is ${VERSION_ID}"
     fi
 
-    local IMAGE_URL="${BASE_URL}/${VERSION_ID}/${IMAGE_NAME}"
-    local SIG_NAME="${IMAGE_NAME}.sig"
-    local SIG_URL="${BASE_URL}/${VERSION_ID}/${SIG_NAME}"
-
+    IMAGE_URL="${BASE_URL}/${VERSION_ID}/${IMAGE_NAME}"
+    SIG_NAME="${IMAGE_NAME}.sig"
+    SIG_URL="${BASE_URL}/${VERSION_ID}/${SIG_NAME}"
     if ! wget ${WGET_ARGS} --spider --quiet "${IMAGE_URL}"; then
         echo "$0: Image URL unavailable: $IMAGE_URL" >&2
         exit 1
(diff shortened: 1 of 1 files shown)


bin/flatcar-install
```diff
@@ -600,6 +632,8 @@ trap 'error_output ; is_modified && wipefs --all --backup "${DEVICE}" ; rm -rf "
 
 if [ -n "${IMAGE_FILE}" ]; then
     install_from_file
+elif [[ "${DOWNLOAD_ONLY}" = 'true' ]]; then
+    download_from_url
 else
     install_from_url
 fi
```
pothos, line 636: I would move the below lines into the `else` branch. Then you can remove the `trap … exit` from the function above and put the `echo …` in the `elif` branch.

bin/flatcar-install
```diff
@@ -538,23 +545,48 @@ function install_from_url() {
...
+        | gpg --batch --trusted-key "${GPG_LONG_ID}" \
+            --verify "${WORKDIR}/${LOCVAR_SIG_NAME}" -
+    then
+        local EEND=( "${PIPESTATUS[@]}" )
+        [ ${EEND[0]} -ne 0 ] && echo "${EEND[0]}: Download of ${LOCVAR_IMAGE_NAME} did not complete" >&2
+        [ ${EEND[1]} -ne 0 ] && echo "${EEND[1]}: Cannot expand ${LOCVAR_IMAGE_NAME} to ${DEVICE}" >&2
+        [ ${EEND[2]} -ne 0 ] && echo "${EEND[2]}: GPG signature verification failed for ${LOCVAR_IMAGE_NAME}" >&2
+        exit 1
+    fi
+    trap - EXIT
```
pothos: See comment below to move this out of the function. Also, it misses the `WORKDIR` removal.

bin/flatcar-install
```diff
@@ -380,59 +382,64 @@ Settings:
...
+            echo "$0: Ignition config file (${IGNITION}) does not exist." >&2
+            exit 1
+        fi
     fi
-fi
 
-if [[ -n "${DRY_RUN}" ]]; then
-    print_settings
-    exit 0
+    if [[ -n "${DRY_RUN}" ]]; then
```
pothos: Is is ok to leave the `print_settings` case out of the `if` block.

jasonbraganza: end the if block above this @pothos?
as in skip until the Ignition check?

bin/flatcar-install
```diff
@@ -341,7 +342,7 @@ VERSION_SUMMARY='Flatcar Container Linux'
 
 WGET_ARGS="--tries 10 --timeout=20 --retry-connrefused"
 
-while getopts "V:B:C:I:d:o:c:e:i:t:b:k:f:nsyvh" OPTION
+while getopts "V:B:C:I:d:o:c:e:i:t:b:k:f:nsyvhD" OPTION
```
pothos: Aiven the ASCII ordering I would move `D` to the other capital letters, also in the `case` block below.

bin/flatcar-install
pothos: yes

bin/flatcar-install
jasonbraganza: Done.

bin/flatcar-install
jasonbraganza: done. dry run excluded from the download flag check block

bin/flatcar-install
```diff
@@ -480,14 +487,14 @@ function install_from_file() {
...
 }
 
-function install_from_url() {
+function prep_url(){
     # Ensure that required executables exist before proceeding
     type -P wget >/dev/null || { echo 'Missing wget!' >&2 ; exit 1 ; }
     type -P gpg >/dev/null || { echo 'Missing gpg!' >&2 ; exit 1 ; }
 
-    local IMAGE_NAME="flatcar_production_image.bin.bz2"
+    LOCVAR_IMAGE_NAME="flatcar_production_image.bin.bz2"
```
margamanterola: Why did you change all the "local VARNAME" statements into "LOCVAR_" statements? If a variable shouldn't be local, then you would remove the local statement, but preceding it with LOCVAR is just confusing.

bin/flatcar-install
```diff
@@ -538,23 +545,48 @@ function install_from_url() {
...
+    prep_url
+    echo "Downloading, writing and verifying ${LOCVAR_IMAGE_NAME}..."
+    if ! wget ${WGET_ARGS} --no-verbose --show-progress -O - "${LOCVAR_IMAGE_URL}" \
+        | tee "${PWD}/${LOCVAR_IMAGE_NAME}" \
+        | gpg --batch --trusted-key "${GPG_LONG_ID}" \
+            --verify "${WORKDIR}/${LOCVAR_SIG_NAME}" -
+    then
+        local EEND=( "${PIPESTATUS[@]}" )
+        [ ${EEND[0]} -ne 0 ] && echo "${EEND[0]}: Download of ${LOCVAR_IMAGE_NAME} did not complete" >&2
+        [ ${EEND[1]} -ne 0 ] && echo "${EEND[1]}: Cannot expand ${LOCVAR_IMAGE_NAME} to ${DEVICE}" >&2
```
margamanterola: This status code makes no sense as there's no expansion to device

bin/flatcar-install
```diff
@@ -538,23 +545,48 @@ function install_from_url() {
...
+   
+    VERSION_SUMMARY+=" ${CHANNEL_ID} ${VERSION_ID}${OEM_ID:+ (${OEM_ID})}"
+
+}
+
+function download_from_url(){
+    prep_url
+    echo "Downloading, writing and verifying ${LOCVAR_IMAGE_NAME}..."
+    if ! wget ${WGET_ARGS} --no-verbose --show-progress -O - "${LOCVAR_IMAGE_URL}" \
+        | tee "${PWD}/${LOCVAR_IMAGE_NAME}" \
```
margamanterola: Given that this command is not doing the decompressing and writing to disk on the fly operation that the other command is doing, you don't need to pipe all the commands.  Instead, you can just tell wget to write the file (i.e use `-O filename` instead of `-O -`). In that case, you don't need tee and you don't need to do magic with PIPESTATUS, just call wget first, and gpg second, printing the corresponding error if something went wrong.

bin/flatcar-install
```diff
@@ -600,14 +628,22 @@ trap 'error_output ; is_modified && wipefs --all --backup "${DEVICE}" ; rm -rf "
...
     install_from_file
+elif [[ "${DOWNLOAD_ONLY}" = 'true' ]]; then
+    download_from_url
+ 
+    rm -rf "${WORKDIR}"
+
+    echo "Success! ${VERSION_SUMMARY} has been downloaded to the current directory"
 else
     install_from_url
+    wait_for_disk
```
margamanterola, line 632: This means that install_from_file doesn't work now. I think we want:

```
if [[ "${DOWNLOAD_ONLY}" = 'true' ]]; then
    download_from_url
else
    if [ -n "${IMAGE_FILE}" ]; then
        install_from_file
    else
        install_from_url
    fi
    wait_for_disk
    write_cloudinit
    write_ignition
    echo "Success! ${VERSION_SUMMARY} is installed on ${DEVICE}"
fi

rm -rf "${WORKDIR}"
trap - EXIT
```

jasonbraganza:
@marga-kinvolk, @pothos thank you both. will work on all you’ve said.

bin/flatcar-install
```diff
@@ -600,14 +628,22 @@ trap 'error_output ; is_modified && wipefs --all --backup "${DEVICE}" ; rm -rf "
 
 if [ -n "${IMAGE_FILE}" ]; then
     install_from_file
+elif [[ "${DOWNLOAD_ONLY}" = 'true' ]]; then
+    download_from_url
+ 
+    rm -rf "${WORKDIR}"
```
```suggestion
    rm -rf "${WORKDIR}"
    trap - EXIT
```
or move these commands out of the if-else blocks (as Marga suggested)

bin/flatcar-install
jasonbraganza: @marga-kinvolk, thank you for this bit. I would have gotten here eventually, but after a *lot* of head scratching :)

bin/flatcar-install
jasonbraganza: prefixes removed.
i was being conservative and assumed they were being declared local, because we might want to use them elsewhere as well and I did not want to muddle and break things I don’t fully understand, yet.
thank you @marga-kinvolk

bin/flatcar-install
jasonbraganza: understood. will do so.

jasonbraganza:
working on all of this. almost done. will update once i am done testing.

jasonbraganza:
d2

@marga-kinvolk @pothos - done. would be grateful for further advice or action.

bin/flatcar-install
```diff
@@ -540,7 +546,27 @@ function install_from_url() {
...
     wget ${WGET_ARGS} --no-verbose -O "${WORKDIR}/${SIG_NAME}" "${SIG_URL}"
+   
+    VERSION_SUMMARY+=" ${CHANNEL_ID} ${VERSION_ID}${OEM_ID:+ (${OEM_ID})}"
+
+}
+
+function download_from_url(){
+    prep_url
+    echo "Downloading, writing and verifying ${IMAGE_NAME}..."
+    if ! wget ${WGET_ARGS} --no-verbose --show-progress -O "${PWD}/${IMAGE_NAME}"  "${IMAGE_URL}"; then
```
pothos: Looking good! I would remove `--show-progress` here or make it optional because it makes a large log output for scripts.

bin/flatcar-install
```diff
@@ -540,7 +546,27 @@ function install_from_url() {
...
+        echo "Could not download ${IMAGE_NAME}"
```
```suggestion
        echo "Could not download ${IMAGE_NAME}" >&2
```

jasonbraganza: oh shoot! fixed. thank you.

bin/flatcar-install
jasonbraganza: understood. will work at it. thank you.

jasonbraganza: just read through flatcar-install -h for inspiration. nothing there is chatty. 
makes more sense like you suggested @pothos, to remove `--show-progress`

bin/flatcar-install
```diff
@@ -515,10 +521,10 @@ function install_from_url() {
...
     fi
 
-    local IMAGE_URL="${BASE_URL}/${VERSION_ID}/${IMAGE_NAME}"
-    local SIG_NAME="${IMAGE_NAME}.sig"
-    local SIG_URL="${BASE_URL}/${VERSION_ID}/${SIG_NAME}"
-
+    IMAGE_URL="${BASE_URL}/${VERSION_ID}/${IMAGE_NAME}"
+    SIG_NAME="${IMAGE_NAME}.sig"
+    SIG_URL="${BASE_URL}/${VERSION_ID}/${SIG_NAME}"
+
```
margamanterola: Nit: this line has some trailing whitespace, can you get rid of it please?

bin/flatcar-install
```diff
@@ -540,7 +546,27 @@ function install_from_url() {
...
     wget ${WGET_ARGS} --no-verbose -O "${WORKDIR}/${SIG_NAME}" "${SIG_URL}"
+   
+    VERSION_SUMMARY+=" ${CHANNEL_ID} ${VERSION_ID}${OEM_ID:+ (${OEM_ID})}"
+
+}
+
+function download_from_url(){
+    prep_url
+    echo "Downloading, writing and verifying ${IMAGE_NAME}..."
+    if ! wget ${WGET_ARGS} --no-verbose -O "${PWD}/${IMAGE_NAME}"  "${IMAGE_URL}"; then
```
margamanterola: Nit: there's two spaces between the params when one is enough, can you get rid of that please?

bin/flatcar-install
```diff
@@ -67,6 +67,7 @@ done
 USAGE="Usage: $0 {{-d <device>}|-s} [options]
 Options:
     -d DEVICE   Install Flatcar Container Linux to the given device.
+    -D          Download Flatcar Container Linux. (will download to a subdirectory in your current directory)
```
margamanterola: The comment here says that it will download to a subdirectory, but the message at the end of the script says that it's the current directory... I think the current directory is actually the right one (I haven't run the script to check).  If that's the case, I'd say the comment should be: "Download a Flatcar Container Linux image to the current directory, without installing it".

margamanterola (commented):
Thanks, this looks much better.  Just one comment about the command documentation and two nits (this is small comments that are not very important).

Once you've solved this, would you please squash your commits into just one and make sure that the commit description has the right information? Thanks.

bin/flatcar-install
jasonbraganza: Please nitpick all you want.  
It helps me a lot as a beginner.
Shows me the difference between how the pros do it and how I as a rank amateur stack up against this.
stripped trailing whitespace.

bin/flatcar-install
jasonbraganza: stripped extraneous whitespace

bin/flatcar-install
jasonbraganza: Gah! thank you for noticing, @marga-kinvolk 
Now I realise how multiple eyes on the code help. (specially from more experienced, wiser owls)
I had wanted to do it this way originally, until i realised otherwise.
Have updated to your text.

jasonbraganza:
Te squash and commit are done.
(lots of thanks to @sayanchowdhury, for helping me out with git hiccoughs)

pothos (approved):
Gave it another test run and it works good, thanks!
A minor thing is that `umask 077` makes the downloaded file restricted to the downloading user which could be reverted afterwards (but not skipped because it's maybe relevant for the verification keys in `/tmp`)

jasonbraganza:
what would you suggest the mask be, @pothos? i could try working on that aspect too

pothos:
I think it's out of scope, but I would say before running `umask 077` the previous umask is saved (`PREV_UMASK=$(umask)`) and then in the download function the umask is restored before `wget` starts (`umask ${PREV_UMASK}`).

jasonbraganza:
@pothos just so i understand correctly, you want the downloaded file to have 755 permissions so the download ought to run with a 022 mask and then set back to 077?

as for scope, well, i want to learn and do, so no worries there :)

pothos:
I meant we can first merge the PR as is if you want :) Yes, the umask is done to limit the temporary gpg folder I think and it's good to keep it, but I would just reset the umask to the previous value before the download starts. This way the user can still control the final permissions by tweaking umask before starting the download script.

jasonbraganza:
@pothos aah that sounds a teensy bit involved, so yes, i’ll wait for the pr to be merged and do this as a separate pr, if it’s ok with you folks :)

pothos:
Yes, thanks again

pothos:
> I think it's out of scope, but I would say before running `umask 077` the previous umask is saved (`PREV_UMASK=$(umask)`) and then in the download function the umask is restored before `wget` starts (`umask ${PREV_UMASK}`).

Here again the suggestion in a better format:
```
diff --git a/bin/flatcar-install b/bin/flatcar-install
index 25a8db4..088f841 100755
--- a/bin/flatcar-install
+++ b/bin/flatcar-install
@@ -41,6 +41,7 @@ find_smallest_unused_disk() {
     echo "$smallestdisk"
 }
 
+PREV_UMASK=$(umask)
 # Everything we do should be user-access only!
 umask 077
 
@@ -552,6 +553,7 @@ function prep_url(){
 function download_from_url(){
     prep_url
     echo "Downloading, writing and verifying ${IMAGE_NAME}..."
+    umask ${PREV_UMASK}
     if ! wget ${WGET_ARGS} --no-verbose -O "${PWD}/${IMAGE_NAME}" "${IMAGE_URL}"; then
         echo "Could not download ${IMAGE_NAME}" >&2
         exit 1
```

jasonbraganza:
Thank you @pothos for showing me that what you were suggesting is in fact, simple :)

Thank you so much @marga-kinvolk and @pothos for guiding and herding me along. This is my first major contribution to an open source project. Feels really fun :)

PR #38, opened 2021-03-12
systemd/network: set predictable alternative ifnames for virtio

pothos:
-    udev/rules.d: replace 79-net-google-compat.rules hardcoding ens4v1   
    With the systemd/udev 247 change of bind/unbind the old rule to set
    ens4v1 as interface name on GCE broke and only eth0 was used now.
    The old name ens4v1 is not even a current predictable name and the
    breakage already appeared on Stable but for compatibility we can
    support this name as alternative interface name instead of
    maintaining the udev rule. For that, add special networkd link
    files for GCE.
-   systemd/network: set predictable alternative ifnames for virtio   
    The virtio interfaces use unpredictable names like eth0 which is
    good as long as there is only one interface. In case there are multiple
    interfaces it makes sense to use predictable names at least as
    alternative names as way of referring to them.

Note: Consider to backport this to 2765.

# How to use/testing done

Tested with
```
{ cat /usr/lib64/systemd/network/98-virtio.link; echo "AlternativeNamesPolicy=database onboard slot path"; echo "AlternativeName=ens4v1"; } | sudo tee /etc/systemd/network/98-virtio.link
```
on QEMU and GCE. The interface name is `eth0` and there are two alternative names `ensX` and `enp0sX` attached to it (X is 4 for GCE, 3 for QEMU in my case) and the old hardcoded `ens4v1` name.


margamanterola (approved):
LGTM.

I find it weird that we need two files for alternate matching, but I did some digging myself and couldn't find a better way either.  Thanks!

PR #41, opened 2021-07-05
udev: add systemd tag for Azure storage devices

pothos:
Without the systemd tag no device units are created for the storage
devices. This causes the /boot automount dependency to be missing.

Add the systemd tag in udev for Azure storage devices, both on
partitions and the full disk.

## How to use/Testing done

See https://github.com/kinvolk/coreos-overlay/pull/1091


ericgraf (approved):
This change worked for me in my environment.

jepio (approved):

PR #44, opened 2021-08-02
add warning to motd when cgroupv1 is in use

jepio:
The warning text is orange and will complement the "taint" warning that systemd v248 reports on such systems. This is part of the work to update to Docker 20.10 and Cgroup v2.

Part of kinvolk/coreos-overlay#931

### DIFF
--- a/scripts/motdgen
+++ b/scripts/motdgen
@@ -22,6 +22,10 @@ if [[ -d "/etc/motd.d" ]]; then
 	cat /etc/motd.d/*.conf 2>/dev/null >> /run/flatcar/motd || true
 fi
 
+if ! ( mount | grep -q cgroup2 ); then
+	echo -e "\e[33mWarning\e[39m: using legacy cgroups" >> /run/flatcar/motd
+fi
+
 if [[ "${GROUP}" =~ "lts" ]]; then
 	echo "FCL LTS documentation: https://kinvolk.io/fcl-lts-docs" >> /run/flatcar/motd
 fi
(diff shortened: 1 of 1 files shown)


scripts/motdgen
```diff
@@ -22,6 +22,10 @@ if [[ -d "/etc/motd.d" ]]; then
 	cat /etc/motd.d/*.conf 2>/dev/null >> /run/flatcar/motd || true
 fi
 
+if ! ( mount | grep -q cgroup2 ); then
+	echo -e "\e[33mWarning\e[39m: using legacy cgroups" >> /run/flatcar/motd
```
pothos, line 26: I guess the docs link would go here?
```suggestion
	echo -e "\e[33mWarning\e[39m: using legacy cgroups, for more details, visit" >> /run/flatcar/motd
	echo "  https://kinvolk.io/docs/flatcar-container-linux/latest/container-runtimes/switching-to-unified-cgroups" >> /run/flatcar/motd
```

jepio: Its a long link (see https://github.com/kinvolk/update_engine/pull/13/files#diff-94291a3f49cb69d1ae2d19d46bd09eda17c876a734961eeb686217b07f864cd3R342) - do you think it should be printed here?

pothos: Yes, people may not even know that the grub file exists

pothos (approved):

scripts/motdgen
jepio: Added the line

PR #46, opened 2021-08-26
flatcar-install uses lbzip2 if present, falls back on bzip2 if not

aniruddha2000:
# Check lbzip2 and set it in global variable

Check if lbzip2 is there or not. If yes then set it to a global variable called BZIP_UTIL or if not then set bzip2 to the global variable and use it for further work.

## Testing done

Added lbzip2 to the custom Flatcar image and checked with `flatcar-install` script `flatcar-install -d /dev/vdb` (with and without the `-f` flag).
```bash
Writing flatcar_production_qemu_image.img.bz2...
Success! Flatcar Container Linux (from flatcar_production_qemu_image.img.bz2) is installed on /dev/vdb
```

### DIFF
--- a/bin/flatcar-install
+++ b/bin/flatcar-install
@@ -5,6 +5,12 @@
 
 set -e -o pipefail
 
+if command -v lbzip2 > /dev/null; then
+    BZIP_UTIL=lbzip2
+else
+    BZIP_UTIL=bzip2
+fi
+
 error_output() {
     echo "Error: return code $? from $BASH_COMMAND" >&2
 }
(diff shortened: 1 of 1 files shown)


pothos (commented):
Looks good, can you squash the commits (and use the message from the second)?

aniruddha2000:
@pothos yup sure!

pothos (approved):
Tested with `sudo ./flatcar-install -s -o "" -C stable` on a QEMU VM and it uses lbzip2 when available and otherwise bzip2, thanks.

PR #52, opened 2021-10-28
Remove logrotate service, timer and config files

aniruddha2000:
# Remove logrotate service, timer and config files
As the package app-admin/logrotate is moving from portage-stable to coreos-overlay. We are moving the service file and the config files to the package files directory in the coreos-overlay. So we no longer need them in this repo.


pothos (commented):
The logrotate.timer file can also go away (https://github.com/flatcar-linux/coreos-overlay/pull/1370#pullrequestreview-792287863)

krnowak:
Should we also remove `systemd/system/multi-user.target.wants/logrotate.timer`? Now it's a dangling symlink, but does the updated logrotate package install the symlink too? If so, we could remove it from this repository too.

krnowak (changes requested):
> Should we also remove `systemd/system/multi-user.target.wants/logrotate.timer`? Now it's a dangling symlink, but does the updated logrotate package install the symlink too? If so, we could remove it from this repository too.

The logrotate package doesn't install the symlink, so I propose dropping it from `init` and do the symlink installation in the updated package, since we have already moved the package to coreos-overlay.

krnowak (changes requested):
The changes look good, thanks. Could you please squash the commits and add the reason for the changes in the commit message? The one you wrote in the PR message sounds good, so you could reuse it.

aniruddha2000:
Sure!

aniruddha2000:
Squashed and updated the commit hash in the init ebuild.

krnowak:
@aniruddha2000 : Sorry for not being clear - I meant adding the following text to the commit message, it explains why we make this change pretty well:

```
As the package app-admin/logrotate is moving from portage-stable
to coreos-overlay. We are moving the service file and the config
files to the package files directory in the coreos-overlay. So we
no longer need them in this repo.
```

Also, please wait with updating the commit hash in the init ebuild - we will want to use the merge commit hash, so you'll get it after this PR is merged. :)

krnowak (approved):
Thanks!

PR #53, opened 2021-11-30
bin: add flatcar-update tool for manual updates or airgapped updates

pothos:
While updates are managed through Nebraska, not every user wants to or
can run their own service. Sometimes it would be more convenient if a
local update command exists that also supports airgapped updates or
helps a user to quickly jump between versions to reproduce an error.

Add a flatcar-update tool (similar to flatcar-install) which users can
run on the machine to force an update to a particular release or local
update payload. On old machines this script can be fetched and
executed, too, and should be CoreOS-compatible. Since update-engine
does no version check, it's also possible to downgrade with the tool.
The tool can disable updates after it ran, making it easy to do manual
updates when no own Nebraska server exists.

Docs PR: https://github.com/flatcar-linux/flatcar-docs/pull/191

## How to use

See `./flatcar-update -h` - depending on whether you run it from a dev image or a release image, set the force key option.

Maybe consider to backport.

## Testing done

A few test runs on a VM.

↓ will do in coreos-overlay

### DIFF
--- a/bin/flatcar-update
+++ b/bin/flatcar-update
@@ -1,0 +1,187 @@
+#!/bin/bash
+set -euo pipefail
+
+opts=$(getopt --name "$(basename "${0}")" --options 'hV:P:DFA' \
+       --longoptions 'help,to-version:,to-payload:,force-dev-key,force-flatcar-key,disable-afterwards' -- "${@}")
+eval set -- "${opts}"
+
+USER_PAYLOAD=
+PAYLOAD=
+VERSION=
+FORCE_DEV_KEY=
+FORCE_FLATCAR_KEY=
+DISABLE_AFTERWARDS=
+
+while true; do
+  case "$1" in
+  -h|--help)
+    echo "Usage: $(basename "${0}") --to-version VERSION [--to-payload FILENAME] [--force-dev-key|--force-flatcar-key|--disable-afterwards]"
+    echo "  Updates Flatcar Container Linux through a temporary local update service on localhost."
+    echo "  The update-engine service will be unmasked (to disable updates again use -A)."
+    echo "  The reboot should be done after applying the update, either manually or through your reboot manager (check locksmithd/FLUO)."
+    echo "  An error will be reported if a previously applied update wasn't booted into yet (you may discard it with 'update_engine_client -reset_status')."
+    echo "  Warning: If you jump between channels, delete any GROUP configured in /etc/flatcar/update.conf for the new defaults to apply."
+    echo "Options:"
+    echo "  -V, --to-version <VERSION>	Updates to the version, by default using the matching release from update.release.flatcar-linux.net"
+    echo "  -P, --to-payload <FILENAME>	Updates to the given update payload file instead of downloading it"
+    echo "  -D, --force-dev-key		Bind-mounts the dev key over /usr/share/update_engine/update-payload-key.pub.pem"
+    echo "  -F, --force-flatcar-key	Bind-mounts the Flatcar release key over /usr/share/update_engine/update-payload-key.pub.pem"
+    echo "  -A, --disable-afterwards	Writes SERVER=disabled to /etc/flatcar/update.conf when done (this overwrites any custom SERVER)"
+    echo
+    echo "Example for updating to the latest Stable release and disabling automatic updates afterwards:"
+    echo '  VER=$(curl -fsSL https://stable.release.flatcar-linux.net/amd64-usr/current/version.txt | grep FLATCAR_VERSION= | cut -d = -f 2)'
+    echo "  $(basename "${0}") -V \$VER -A"
+    exit 1
+    ;;
+  -V|--to-version)
+    shift
+    VERSION="$1"
+    ;;
+  -P|--to-payload)
+    shift
+    PAYLOAD="$1"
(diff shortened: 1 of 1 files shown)


bin/flatcar-update
```diff
@@ -0,0 +1,173 @@
...
+PAYLOAD=
+VERSION=
+FORCE_DEV_KEY=
+FORCE_FLATCAR_KEY=
+DISABLE_AFTERWARDS=
+
+while true; do
+  case "$1" in
+  -h|--help)
+    echo "Usage: $(basename "${0}") --to-version VERSION [--to-payload FILENAME] [--force-dev-key|--force-flatcar-key|--disable-afterwards]"
```
tormath1, line 18: what about adding a simple common example ? For one who just want to know how to update flatcar and avoid passing through the options in the help ?

bin/flatcar-update
```diff
@@ -0,0 +1,173 @@
...
+    DISABLE_AFTERWARDS=1
+    ;;
+  --)
+    shift
+    break;;
+  esac
+  shift
+done
+
+if [ "$VERSION" = "" ]; then
```
tormath1: Maybe just for consistency we can use 
```suggestion
if [ "${VERSION}" = "" ]; then
```

bin/flatcar-update
```diff
@@ -0,0 +1,173 @@
...
+if [ "$STATUS" != "" ]; then
+  echo "Error: a previously downloaded update wasn't applied yet, you can discard it with 'update_engine_client -reset_status'" > /dev/stderr; exit 1
+fi
+
+touch /etc/flatcar/update.conf
+PREV_SERVER=$(grep '^SERVER=' /etc/flatcar/update.conf || true)
+sed -i "/SERVER=.*/d" /etc/flatcar/update.conf
+
+echo "SERVER=http://localhost:9090/update" >> /etc/flatcar/update.conf
+BOARD=$({ grep -m 1 BOARD= /usr/share/coreos/release || true ; } | cut -d = -f 2-)
```
tormath1, line 111: Any reason to grep onto `/usr/share/coreos/release` ? On legacy CoreOS `/etc/os-release` does not exist ?

bin/flatcar-update
```diff
@@ -0,0 +1,173 @@
...
+if [ "$BOARD" = "" ]; then
+  echo "Error: could not find board from /usr/share/coreos/release" > /dev/stderr ; exit 1
+fi
+
+if [ "$PAYLOAD" = "" ]; then
+  PAYLOAD="/var/tmp/update_payload"
+  rm -f "$PAYLOAD"
+  echo "Downloading update payload..."
+  curl -fsSL -o "$PAYLOAD" --retry-delay 1 --retry 60 --retry-connrefused --retry-max-time 60 --connect-timeout 20 "https://update.release.flatcar-linux.net/${BOARD}/${VERSION}/flatcar_production_update.gz"
+  echo "When restarting after an error you may reuse it with '--to-payload $PAYLOAD'"
```
tormath1, line 130: We could also download the the `*.sha256` file to check the download integrity:
```
sha256sum -c flatcar_production_update.gz.sha256
flatcar_production_update.gz: OK
```

Though it does not really work this way on latest updates since it's an absolute path - see: https://update.release.flatcar-linux.net/amd64-usr/3066.0.0/flatcar_production_update.gz.sha256 so we might need to compare "manually".

bin/flatcar-update
tormath1: From your documentation example:
```
sudo flatcar-update --to-version 2905.2.6 --disable-afterwards
```

bin/flatcar-update
pothos: Thanks, done

bin/flatcar-update
pothos: done

bin/flatcar-update
pothos: Thanks, done

bin/flatcar-update
pothos: The read-only file on the active `/usr` partition is the best bet, the user can overwrite the file under `/etc`

bin/flatcar-update
pothos: wait, got the hex vs base64 wrong

bin/flatcar-update
tormath1: Oh I see. We still have the `/usr/lib/os-release` which is the source of `/etc/os-release` following `systemd-tmpfiles` configuration.
They basically holds the same configuration - it's just that we could avoid reintroducing `coreos` occurrences mention. :)

bin/flatcar-update
pothos: done now

bin/flatcar-update
pothos: The `/usr/share/coreos/release`/`/usr/share/flatcar/release` file is the file for update-engine, the `coreos` symlink is there for backwards compatibility, I could add a check which one to use but I didn't expect the symlink to go soon

tormath1: Oh ok understood, then let's keep things in this way. :)

tormath1 (approved):
LGTM, thanks for this :)

PR #55, opened 2021-12-09
networkd: prevent networkd interference with kubenet cbr0 interface

pothos:
The Kubernetes Kubenet CNI bridge was getting the default networkd
configuration applied which tries to attach an IP address through DHCP.
This meant that the interface was stuck in this "configuring" state by
networkd and this is also a potential source of disruption as reported
in similar cases.

Set the cbr0 interface to be excluded from networkd (unmanaged) because
it is set up manually by kubenet and not through DHCP.

## Testing done

I copied the new config to `/etc/systemd/network/cbr.network` and saw it take effect

↑ will do in coreos-overlay

### DIFF
--- a/systemd/network/kubenet.network
+++ b/systemd/network/kubenet.network
@@ -1,0 +1,7 @@
+# Exclude Kubernetes Kubenet CNI bridge from DHCP by default
+
+[Match]
+Name=cbr0
+
+[Link]
+Unmanaged=yes


systemd/network/kubenet.network
```diff
@@ -0,0 +1,7 @@
+# Exclude Kubernetes Kubenet CNI bridge from DHCP by default
+
+[Match]
+Name=cbr0
```
krnowak, line 4: Should this be `cbr*` maybe?

krnowak (commented):
Is this potentially being superseded by #56?

systemd/network/kubenet.network
pothos: There is only a singe bridge interface and I think that other interface names like `cbroutemyinterfacewantstouseitsownconfig7` would want to be able to have their own config instead of being included in this one and having to fight over which wins.

krnowak: Alright, let's have `cbr0` then. Probably having something like `cbr[0-9]+` would be better, but I have just read that shell-style globs only are allowed.

krnowak (approved):

PR #57, opened 2021-12-16
configs/tmpfiles.d: ensure /run/xtables.lock exists

### LINKED ISSUE #578: /run/xtables.lock file missing after upgrading to stable 3033.2.0
## Description

After rebooting to Flatcar stable 3033.2.0, `/run/xtables.lock` file is no longer present on the host due to iptables upgrade to use nftables backend.

## Impact

In my case, I run kubelet in a Docker container, so I have to bind-mount `/run/xtables.lock` from host into kubelet container to avoid concurrent modification of iptables rules. Because of this and lack of this file after the reboot, Kubernetes node failed to come back after a reboot. Fortunately FLUO did not proceed with upgrading next nodes.

## Workaround

The issue can be at least temporarily workaround by executing the following commands before the reboot for upgrade:
```sh
sudo mkdir -p /etc/systemd/system/docker.service.d

cat <<EOF | sudo tee /etc/systemd/system/docker.service.d/override.conf
[Unit]
Requries=systemd-tmpfiles-setup.service
After=systemd-tmpfiles-setup.service
EOF

cat << EOF | sudo tee /etc/tmpfiles.d/xtables-workaround.conf
f       /run/xtables.lock       0755    root    root    -       -
EOF
```

## Additional information

Opening mainly for documentation purposes requested by @t-lo.

pothos:
The nftables update which included using the nftables compat backend
for the iptables binaries instead of xtables on the host resulted in
the lock file not being created anymore automatically. The lock file is
still required because the xtables backend doesn't go away and is used
by containers and, possibly, by invoking the legacy binaries on the
host (we ship them for easy access to the xtables lists).

Use a systemd-tmpfile directive to create the xtables lock file which,
e.g., gets bind-mounted to containers for coordination of xtables
modifications.

## How to use

Fixes https://github.com/flatcar-linux/Flatcar/issues/578

## Testing done

File got created at bootup in a VM. After file deletion the same file got created again by `iptables-legacy -L`

↑ in coreos-overlay

### DIFF
--- a/configs/tmpfiles.d/xtables-lock.conf
+++ b/configs/tmpfiles.d/xtables-lock.conf
@@ -1,0 +1,4 @@
+# This file gets still created by iptables-legacy but since
+# iptables-nft is used now by default it doesn't but some containers
+# still have to bind mount it for legacy iptables operations.
+f       /run/xtables.lock       0600    -    -    -       -


configs/tmpfiles.d/xtables-lock.conf
```diff
@@ -0,0 +1 @@
+f       /run/xtables.lock       0755    root    root    -       -
```
invidian: I'm not sure if the ownership and permissions on the lock file are correct.

pothos: Good catch, will have to compare it with an old version

pothos: should be `0600` (root is correct but I think since it's the default we can omit it)

configs/tmpfiles.d/xtables-lock.conf
```diff
@@ -0,0 +1 @@
+f       /run/xtables.lock       0600    -    -    -       -
```
invidian, line 4: Should we have some tests or a comment why this is needed?

invidian (approved):

configs/tmpfiles.d/xtables-lock.conf
pothos: Thanks, added a comment

pothos:
Created a flatcar-3033-backport branch for current Stable

PR #60, opened 2022-02-11
systemd/system: add oem cloud init service

tormath1:
these services, based on the ignition execution result will enable or
not cloudinit based on OEM.

<hr>

With ignition 3, there is no more "default" configuration (`default.ign`) on each OEM partition. So we need to add a mechanism to conditionally start cloudinit on the supported OEM.

This mechanism is based on the values of the `/etc/.ignition-result.json` result file produced by IgnitionV3.

Tested through: https://github.com/flatcar-linux/coreos-overlay/pull/1249

### DIFF
--- a/systemd/system/oem-cloudinit.service
+++ b/systemd/system/oem-cloudinit.service
@@ -1,0 +1,11 @@
+[Unit]
+Description=Run cloudinit
+
+[Service]
+EnvironmentFile=/var/run/ignition.env
+Type=oneshot
+ExecCondition=/usr/bin/bash -xc 'OEMS=(aws gce rackspace-onmetal azure cloudsigma packet vmware digitalocean); echo $${OEMS[*]} | tr " " "\n" | grep -q -x -F "${OEM_ID}"'
+ExecStart=/usr/bin/bash -xc '/usr/bin/coreos-cloudinit --oem="$(if [ "${OEM_ID}" = aws ]; then echo ec2-compat; else echo "${OEM_ID}" ; fi)"'
+
+[Install]
+WantedBy=multi-user.target
--- a/systemd/system/enable-oem-cloudinit.service
+++ b/systemd/system/enable-oem-cloudinit.service
@@ -1,0 +1,13 @@
+[Unit]
+Description=Enable cloudinit
+
+[Service]
+Type=oneshot
+StandardInput=file:/etc/.ignition-result.json
+ExecCondition=/usr/bin/jq -e '.userConfigProvided == false'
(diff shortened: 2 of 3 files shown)


systemd/system/oem-cloudinit.service
```diff
@@ -0,0 +1,11 @@
+[Unit]
+Description=Run cloudinit
+
+[Service]
+EnvironmentFile=/var/run/ignition.env
+Type=oneshot
+ExecCondition=/usr/bin/bash -xc 'OEMS=(ec2-compat gce rackspace-onmetal azure cloudsigma packet vmware digitalocean); [[ "$${OEMS[*]}" =~ "${OEM_ID}" ]]'
+ExecStart=/usr/bin/coreos-cloudinit --oem="${OEM_ID}"
```
pothos: This here should use `ec2-compat` as argument for `coreos-cloudinit` and not `ec2` which is the OEM_ID

pothos: With Ignition v3 the OEM_ID value inside the `/var/run/ignition.env` file will be `aws`, I suggest to put `aws` in the OEMS list and then change the ExecStart line to pass `ec2-compat` in case `aws` is the OEM_ID value, and otherwise pass it as is.

pothos:
```suggestion
ExecCondition=/usr/bin/bash -xc 'OEMS=(aws gce rackspace-onmetal azure cloudsigma packet vmware digitalocean); echo $${OEMS[*]} | tr ' ' '\n' | grep -q -x -F "${OEM_ID}"'
ExecStart=/usr/bin/bash -xc '/usr/bin/coreos-cloudinit --oem="$(if [ "${OEM_ID}" = aws ]; then echo ec2-compat; else echo "${OEM_ID}" ; fi)"'
```

I also noticed that the `=~` matching also worked for substrings, here I used grep to match the full string of the entry.

systemd/system/oem-cloudinit.service
```diff
@@ -0,0 +1,11 @@
+[Unit]
+Description=Run cloudinit
+
+[Service]
+EnvironmentFile=/var/run/ignition.env
+Type=oneshot
+ExecCondition=/usr/bin/bash -xc 'OEMS=(aws gce rackspace-onmetal azure cloudsigma packet vmware digitalocean); echo $${OEMS[*]} | tr ' ' '\n' | grep -q -x -F "${OEM_ID}"'
```
pothos:
```suggestion
ExecCondition=/usr/bin/bash -xc 'OEMS=(aws gce rackspace-onmetal azure cloudsigma packet vmware digitalocean); echo $${OEMS[*]} | tr " " "\n" | grep -q -x -F "${OEM_ID}"'
```

pothos (approved):

PR #61, opened 2022-02-16
systemd: disable foreign route management

pothos:
- systemd: disable foreign route management
    
    While systemd-networkd follows the principle of a declarative network
    configuration and thus needs a way to ensure that unwanted routes or
    routing policy rules get discarded, the interfacing with procedural
    network management from CNIs like Cilium is limited, so that when the
    interface is set to "unmanaged" through a networkd unit, any routing
    policies there would also be ignored and discarded unless they would
    be defined for a new unit for a dummy network interface. This means
    the only option left is to disable the discarding of foreign rules
    globally.
    
    Set the default for ManageForeignRoutes and
    ManageForeignRoutingPolicyRules to "no" to ensure that we don't
    interfere with the network management of the CNIs. Users that rely on
    the setting can still enable it again but only through a drop-in
    under /etc/systemd/networkd.conf.d/ because this here is a drop-in
    already that takes precedence over the top config file.
    
    See https://github.com/cilium/cilium/issues/18706
    and https://github.com/flatcar-linux/Flatcar/issues/620
    Replaces https://github.com/flatcar-linux/coreos-overlay/pull/1622

- systemd: move files to init repo as unified location
    
    The systemd drop-in files should only be in one repository, and using
    init is better than baselayout for that (baselayout is also used for
    the SDK).

## How to use

Together with baselayout PR

## Testing done

See coreos-overlay PR

↑ TODO in coreos-overlay

### DIFF
--- a/systemd/resolved.conf.d/10-disable-llmnr.conf
+++ b/systemd/resolved.conf.d/10-disable-llmnr.conf
@@ -1,0 +1,2 @@
+[Resolve]
+LLMNR=false
--- a/Makefile
+++ b/Makefile
@@ -24,16 +24,14 @@ install:
 	install -m 755 bin/* $(DESTDIR)/usr/bin
 	ln -sf flatcar-install $(DESTDIR)/usr/bin/coreos-install
 	install -m 755 scripts/* $(DESTDIR)/usr/lib/flatcar
-	install -m 644 systemd/network/* $(DESTDIR)/usr/lib/systemd/network
-	install -m 755 systemd/system-generators/* \
-		$(DESTDIR)/usr/lib/systemd/system-generators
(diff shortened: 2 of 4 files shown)


systemd/resolved.conf.d/10-disable-llmnr.conf
```diff
@@ -0,0 +1,3 @@
+[Resolve]
+LLMNR=false
+
```
krnowak: Trailing empty line.

krnowak (approved):
Looks good.

systemd/resolved.conf.d/10-disable-llmnr.conf
pothos: I guess this was there before? I just moved this folder here.

PR #62, opened 2022-02-22
add init.wrapper to force switch to legacy cgroups after initramfs

jepio:
# add init.wrapper to force switch to legacy cgroups after initramfs

init.wrapper attempts to disable cgroup controllers, unmount cgroup2,
and then inject (using a bind mount) additional kernel command line
arguments. These kernel parameters are used by systemd to proceed to
mount legacy cygroups.

## How to use

To be used with https://github.com/flatcar-linux/bootengine/pull/35

## Testing done

### DIFF
--- a/sbin/init.wrapper
+++ b/sbin/init.wrapper
@@ -1,0 +1,22 @@
+#!/bin/bash
+
+echo "(flatcar) runtime switch to legacy cgroups"
+
+CTRL_FILE=/sys/fs/cgroup/cgroup.subtree_control
+if [[ -f "${CTRL_FILE}" ]]; then
+  find /sys/fs/cgroup -type d -delete 2>/dev/null
+  while read ctrl; do
+    echo "-${ctrl}" >"${CTRL_FILE}"
+  done <"${CTRL_FILE}"
+fi
+umount /sys/fs/cgroup
+
+mkdir -p /tmp
+read cmdline </proc/cmdline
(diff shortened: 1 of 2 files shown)


sbin/init.wrapper
```diff
@@ -0,0 +1,21 @@
...
+if [[ -f "${CTRL_FILE}" ]]; then
+  find /sys/fs/cgroup -type d -delete 2>/dev/null
+  while read ctrl; do
+    echo "-${ctrl}" >"${CTRL_FILE}"
+  done <"${CTRL_FILE}"
+fi
+umount /sys/fs/cgroup
+
+read cmdline </proc/cmdline
+printf "%s %s\n" "$cmdline" "systemd.unified_cgroup_hierarchy=0 systemd.legacy_systemd_cgroup_controller=1" >/tmp/cmdline
```
pothos, line 16:
```suggestion
mkdir -p /tmp
printf "%s %s\n" "$cmdline" "systemd.unified_cgroup_hierarchy=0 systemd.legacy_systemd_cgroup_controller=1" >/tmp/cmdline
```
What is `/tmp` at this moment -  a plain directory on the rootfs? It may not exist then, or? I would add `mkdir -p /tmp` first. Other options could be using `/dev/shm/` instead if that is available even.

jepio: Its a plain directory on the rootfs, and before I added the "rm /tmp/cmdline" you could still find the file if you did `mount --bind / /mnt; cat /mnt/tmp/cmdline`. I'll add the mkdir to be defensive.

pothos (approved):

PR #63, opened 2022-03-02
sbin/init.wrapper: remove processes from subcgroups before deleting

jepio:
# sbin/init.wrapper: remove processes from subcgroups before deleting

In an attempt to prevent the cgroup from being busy when we unbind controllers
from it.

Will be cherry-picked to flatcar-3139.

## How to use

## Testing done

http://jenkins.infra.kinvolk.io:8080/job/os/job/manifest/5007/cldsv/

### DIFF
--- a/sbin/init.wrapper
+++ b/sbin/init.wrapper
@@ -4,10 +4,25 @@ echo "(flatcar) runtime switch to legacy cgroups"
 
 CTRL_FILE=/sys/fs/cgroup/cgroup.subtree_control
 if [[ -f "${CTRL_FILE}" ]]; then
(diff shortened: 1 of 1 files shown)


krnowak (approved):
Hopefully the pids of the processes spawned during this cleanup won't end up in any of the cgroup.procs that we are trying to clean up.

jepio:
This loop:
```
  find /sys/fs/cgroup/*/ -name cgroup.procs -exec grep . {} \; | \
  while read pid; do
    echo "$pid" >/sys/fs/cgroup/cgroup.procs 2>/dev/null
  done
```
also covers "self", thats accounted for :)

PR #65, opened 2022-03-15
systemd: add helper service to start sysext services

pothos:
As systemd-sysext images are not really meant for system services
(that is the role of portable services) the services that they define
don't get started even if they include the symlinks to do so.
Add a helper service that starts the new services by retriggering the
socket, timers and multi-user targets as these are the targets a user
    would normally use to enable units. We can add more as needed.

## How to use

`systemctl start systemd-sysext` or booting when this service is enabled will now start the services defined in the sysext image that is loaded (given they use the multi-user or sockets targets)

## Testing done

the above with a converted torcx image (plus the regular kola build tests as linked in the coreos-overlay PR)

↑ in coreos-overlay

### DIFF
--- a/systemd/system/ensure-sysext.service
+++ b/systemd/system/ensure-sysext.service
@@ -1,0 +1,11 @@
+[Unit]
+BindsTo=systemd-sysext.service
+After=systemd-sysext.service
+DefaultDependencies=no
+[Service]
+Type=oneshot
+RemainAfterExit=yes
+ExecStart=/usr/bin/systemctl daemon-reload
+ExecStart=/usr/bin/systemctl restart --no-block sockets.target timers.target multi-user.target
+[Install]
+WantedBy=sysinit.target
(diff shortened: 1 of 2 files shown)


systemd/system/ensure-sysext.service
```diff
@@ -0,0 +1,10 @@
+[Unit]
+BindsTo=systemd-sysext.service
+After=systemd-sysext.service
+DefaultDependencies=no
+[Service]
+Type=oneshot
+RemainAfterExit=yes
+ExecStart=/bin/bash -c "systemctl daemon-reload; systemctl restart --no-block sockets.target multi-user.target"
```
jepio: Would this work?
```suggestion
ExecStart=systemctl daemon-reload
ExecStart=systemctl restart --no-block sockets.target multi-user.target
```

krnowak (approved):

PR #66, opened 2022-04-01
systemd/system: run update-ssh-keys once after Ignition

### LINKED ISSUE #699: Alpha: ignition v3 should write SSH pub keys to authorized_keys.d
## Description

When updating to the new Ignition version the ebuild file set `distro.writeAuthorizedKeysFragment=false` to write the SSH pub keys to `.ssh/authorized_keys` directly instead of `.ssh/authorized_keys.d/flatcar-ignition` as before.

## Impact

The keys will get lost when update-ssh-keys rewrites the `.ssh/authorized_keys` file

Didn't try to reproduce it, was just reading the code

## Expected behavior

Ignition used to have similar code to what `update-ssh-keys` does in `internal/authorized_keys_d/authorized_keys_d.go` but now that code is gone because Fedora CoreOS uses the internal mechanism of SSH to read from this directory but that mechanism is not used in Flatcar. Therefore, just enabling `writeAuthorizedKeysFragment` is not enough either, because in addition we have to call `update-ssh-keys` after Ignition wrote the files to `authorized_keys.d/ignition`. 

## Additional information

discovered when reading https://github.com/flatcar-linux/coreos-overlay/pull/1784/files

pothos:
The new Ignition version dropped support for creating the
authorized_keys file alongside the authorized_keys.d entry and can only
write one of them.
Call update-ssh-keys once after Ignition ran, for each user that has
the authorized_keys.d folder.

Fixes https://github.com/flatcar-linux/Flatcar/issues/699

## How to use

Test with distro.writeAuthorizedKeysFragment set to the defaul when compiling Ignition

## Testing done

Tested the new unit on a VM that booted with keys from Ignition

↑ in coreos-overlay

### DIFF
--- a/systemd/system/update-ssh-keys-after-ignition.service
+++ b/systemd/system/update-ssh-keys-after-ignition.service
@@ -1,0 +1,13 @@
+[Unit]
+Description=Run update-ssh-keys once after Ignition
+
+[Service]
+Type=oneshot
+StandardInput=file:/etc/.ignition-result.json
+ExecCondition=/usr/bin/jq -e '.userConfigProvided == true'
+ExecCondition=/usr/bin/jq -e '.provisioningBootID | gsub(\"-\"; \"\") | . == $id' --arg id %b
+ExecStart=/bin/bash -c 'ret=0; for nameandhomedir in $(cut -d : -f 1,6 /etc/passwd); do name=$(echo "$nameandhomedir"| cut -d : -f 1); homedir=$(echo "$nameandhomedir"| cut -d : -f 2); if [ -d "$homedir"/.ssh/authorized_keys.d ]; then update-ssh-keys -u "$name" || ret=1; fi; done; exit $ret'
+RemainAfterExit=true
+
+[Install]
+WantedBy=multi-user.target
--- a/systemd/system/multi-user.target.wants/update-ssh-keys-after-ignition.service
+++ b/systemd/system/multi-user.target.wants/update-ssh-keys-after-ignition.service
@@ -1,0 +1,1 @@
+../update-ssh-keys-after-ignition.service


systemd/system/update-ssh-keys-after-ignition.service
```diff
@@ -0,0 +1,13 @@
+[Unit]
+Description=Run update-ssh-keys once after Ignition
+
+[Service]
+Type=oneshot
+StandardInput=file:/etc/.ignition-result.json
+ExecCondition=/usr/bin/jq -e '.userConfigProvided == true'
+ExecCondition=/usr/bin/jq -e '.provisioningBootID | gsub(\"-\"; \"\") | . == $id' --arg id %b
+ExecStart=/bin/bash -c 'ret=0; for name in $(cut -d : -f 1 /etc/passwd); do if [ -d "$(eval echo "~$name")"/.ssh/authorized_keys.d ]; then update-ssh-keys -u "$name" || ret=1; fi; done; exit $ret'
```
krnowak: Up to you if you want to use this suggestion, but would this work too?

```suggestion
ExecStart=/bin/bash -c 'ret=0; for name in $(cut -d : -f 1 /etc/passwd); do if [ -d "$(getent passwd "$name" | cut -d : -f 6)"/.ssh/authorized_keys.d ]; then update-ssh-keys -u "$name" || ret=1; fi; done; exit $ret'
```

krnowak (approved):
I suppose that no ordering is necessary for the `update-ssh-keys-after-ignition.service` unit file, because this stuff will be executed after switching the root, right?

systemd/system/update-ssh-keys-after-ignition.service
pothos: Maybe we can loop over home directories directly then?

pothos: Updated it

pothos: Tested by copying the unit file to a qemu vm

PR #68, opened 2022-04-12
systemd/system/ensure-sysext: skip unit if systemd-sysext is skipped

pothos:
The unit failed to execute when the dependency systemd-sysext.service
was skipped because it's not needed. Normally this is harmless but it
still triggers a OnFailure event as reported in
https://github.com/flatcar-linux/Flatcar/issues/710

Use the same logic for skipping execution as done in
systemd-sysext.service - we can't use "systemctl is-enabled" because
Flatcar's inbuilt unit symlinks are not reported as "enabled".

## How to use

Test that the built image doesn't contain a log about the unit having failed because of a dependency not being run

## Testing done

Works as expected: `ensure-sysext.service was skipped because all trigger condition checks failed`

↑ TODO in coreos-overlay

### DIFF
--- a/systemd/system/ensure-sysext.service
+++ b/systemd/system/ensure-sysext.service
@@ -2,6 +2,12 @@
 BindsTo=systemd-sysext.service
 After=systemd-sysext.service
 DefaultDependencies=no
+# Keep in sync with systemd-sysext.service
+ConditionDirectoryNotEmpty=|/etc/extensions
+ConditionDirectoryNotEmpty=|/run/extensions
+ConditionDirectoryNotEmpty=|/var/lib/extensions
+ConditionDirectoryNotEmpty=|/usr/local/lib/extensions
+ConditionDirectoryNotEmpty=|/usr/lib/extensions
 [Service]
 Type=oneshot
 RemainAfterExit=yes


systemd/system/ensure-sysext.service
```diff
@@ -2,6 +2,12 @@
 BindsTo=systemd-sysext.service
 After=systemd-sysext.service
 DefaultDependencies=no
+# Keep in sync with systemd-sysext.service
```
krnowak, line 5: Hrm. This surely will be forgotten to be done on systemd update. I think we will need a checklist in `sys-apps/systemd` in overlay with a list of things to do on systemd update. Could you add something like `CHECKLIST.md` there?

pothos: Yes, or we merge these actions as ExecStartPost lines into the sysext service as suggested by @jepio - I thought a separate unit would be better, once for the user to know that it does something not upstream, and also in case it fails to run properly.

krnowak: I'm fine with keeping it as a separate unit test as long as we have some way of reminding ourselves that there are some extra actions to be done when `systemd-sysext.service` changes.

krnowak (approved):

PR #69, opened 2022-04-14
systemd-sysext.service.d: add drop-in for OEM partition mount actions

pothos:
When systemd-sysext sets up the new /usr mount, the OEM partition
remains mounted but the mount point is not visible anymore. This breaks
access to grub.cfg, oem-release, and any OEM tools.
Add a drop-in for the systemd-sysext service to first unmount the OEM
partition to prevent stale mounts from piling up that cause the
usr-share-oem.mount unit to fail on stop, and then mount the OEM
partition again when systemd-sysext did its work. At the moment a
"ExecStopPre" action does not exist in systemd and luckily is not
needed because systemd-sysext removes the OEM mount together with the
overlay mounts.  The lazy unmount makes it possible to unmount while a
    process from the OEM tools is running.

## How to use

Check whether the OEM remount interferes with the OEM tool startup

## Testing done

Manually restarted systemd-sysext on Azure (process kept running), on GCE (couldn't test because GCE is broken at the moment, see Jeremi's PRs for the fix) and on VMware ESX (process kept running). The remount on boot was finished before the OEM services started.

↑ TODO in c-o

### DIFF
--- a/systemd/system/systemd-sysext.service.d/10-mount-oem-partition.conf
+++ b/systemd/system/systemd-sysext.service.d/10-mount-oem-partition.conf
@@ -1,0 +1,14 @@
+# Systemd-sysext removes all mounts under /usr and we need to restore the OEM partition mount.
+# Note: This does not cover the rare case of a user-supplied OEM tmpfs for PXE boots,
+#       the user needs to copy the files from the OEM partition to the rootfs via Ignition
+#       or embed them into Ignition.
+#       In the future we can try to move the OEM mount point under /mnt or /run and have
+#       /usr/share/oem be a symlink to it.
+[Unit]
+# Needed for oem-gce-enable-oslogin.service at the moment but as a precaution also done for others in case these get changed to start earlier
+# (from coreos-overlay:coreos-base/oem-*/files/units/*service)
+Before=oem-gce-enable-oslogin.service oem-gce.service waagent.service amazon-ssm-agent.service vmtoolsd.service nvidia.service
+[Service]
+ExecStartPre=/usr/bin/systemctl stop usr-share-oem.mount
+ExecStartPost=/usr/bin/systemctl start usr-share-oem.mount
+ExecStopPost=/usr/bin/systemctl start usr-share-oem.mount
--- a/systemd/system/usr-share-oem.mount
+++ b/systemd/system/usr-share-oem.mount
@@ -14,3 +14,4 @@ What=/dev/disk/by-label/OEM
 Where=/usr/share/oem
 Options=nodev
 Type=auto
+LazyUnmount=yes


systemd/system/systemd-sysext.service.d/10-mount-oem-partition.conf
```diff
@@ -0,0 +1,12 @@
+[Unit]
+Before=oem-gce.service waagent.service amazon-ssm-agent.service vmtoolsd.service nvidia.service
```
jepio: I don't understand why this is necessary - systemd-sysext is a `Before=sysinit.target` service so these should already be ordered correctly.

pothos: Yes, it's not necessary at the moment, it's more a precaution in case `DefaultDependencies=no` would be used in one of them.

pothos: I thought it would be good to have even if this workaround won't be needed for too long hopefully. I can add a comment.

krnowak: Maybe you could also document how did you come up with this list?

systemd/system/systemd-sysext.service.d/10-mount-oem-partition.conf
```diff
@@ -0,0 +1,13 @@
...
+[Service]
+ExecStartPre=/usr/bin/systemctl stop usr-share-oem.mount
+ExecStartPost=/usr/bin/systemctl start usr-share-oem.mount
+ExecStopPost=/usr/bin/systemctl start usr-share-oem.mount
+# Systemd-sysext removes all mounts under /usr and we need to restore the OEM partition mount.
+# Note: This does not cover the rare case of a user-supplied OEM tmpfs for PXE boots,
+#       the user needs to copy the files from the OEM partition to the rootfs via Ignition
+#       or embed them into Ignition.
+#       In the future we can try to move the OEM mount point under /mnt or /run and have
+#       /usr/share/oem be a symlink to it.
```
krnowak, line 6: I'd move it to the top of the file.

systemd/system/systemd-sysext.service.d/10-mount-oem-partition.conf
pothos: Actually it is required for `oem-gce-enable-oslogin.service` to prevent having it fail.

Yes, added `coreos-overlay:coreos-base/oem-*/files/units/*service` as a hint

systemd/system/systemd-sysext.service.d/10-mount-oem-partition.conf
pothos: done

krnowak (approved):

PR #72, opened 2022-06-08
config: add kubernetes configuration

tormath1:
With Kubernetes 1.24 the dockershim component has been removed from the kubelet.
Docker was automatically loading the `br_netfilter` module so we need to
load the module automatically with `systemd-modules-load`.

See also: https://kubernetes.io/docs/setup/production-environment/container-runtimes/#forwarding-ipv4-and-letting-iptables-see-bridged-traffic

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,7 +21,8 @@ install:
 		$(DESTDIR)/usr/lib/systemd/system-generators \
 		$(DESTDIR)/usr/lib/tmpfiles.d \
 		$(DESTDIR)/etc/env.d \
-		$(DESTDIR)/usr/share/ssh
+		$(DESTDIR)/usr/share/ssh \
+		$(DESTDIR)/usr/lib/modules-load.d
 	install -m 755 bin/* $(DESTDIR)/usr/bin
 	install -m 755 sbin/* $(DESTDIR)/usr/sbin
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -21,14 +21,16 @@ install:
...
+		$(DESTDIR)/usr/share/ssh \
+		$(DESTDIR)/usr/lib/modules-load.d
 	install -m 755 bin/* $(DESTDIR)/usr/bin
 	install -m 755 sbin/* $(DESTDIR)/usr/sbin
 	ln -sf flatcar-install $(DESTDIR)/usr/bin/coreos-install
 	install -m 755 scripts/* $(DESTDIR)/usr/lib/flatcar
 	install -m 644 udev/rules.d/* $(DESTDIR)/lib/udev/rules.d
 	install -m 755 udev/bin/* $(DESTDIR)/lib/udev
 	install -m 644 configs/editor.sh $(DESTDIR)/etc/env.d/99editor
+	install -m 644 configs/k8s.conf $(DESTDIR)/usr/lib/modules-load.d/
```
pothos:
```suggestion
	install -m 644 configs/modules-load.d/* $(DESTDIR)/usr/lib/modules-load.d/
```
We can use a subfolder for future additions.

tormath1: Excellent suggestion yes.

dongsupark (approved):

PR #74, opened 2022-06-13
Add an option to create the UEFI flatcar boot entry

jbguerraz:
# flatcar-install: Add an option to create the UEFI flatcar boot entry

## How to use

flatcar-install -u ...

## Testing done

installed a node

- [ ] flatcar-install: Add an option to create the UEFI flatcar boot entry

### DIFF
--- a/bin/flatcar-install
+++ b/bin/flatcar-install
@@ -105,6 +105,7 @@ Options:
     -k KEYFILE  Override default GPG key for verifying image signature.
     -f IMAGE    Install unverified local image file to disk instead of fetching.
     -n          Copy generated network units to the root partition.
+    -u          Create UEFI boot entry
     -y          Dry-run.  Run some checks and print option settings.
     -v          Super verbose, for debugging.
     -h          This ;-).
@@ -395,10 +396,11 @@ CLOUDINIT=
 IMAGE_FILE=
 KEYFILE=
 VERSION_SUMMARY='Flatcar Container Linux'
+CREATE_UEFI=
 
 WGET_ARGS="--tries 10 --timeout=20 --retry-connrefused"
 
-while getopts "V:B:C:DI:d:o:c:e:i:t:b:k:f:nsyvh" OPTION
+while getopts "V:B:C:DI:d:o:c:e:i:t:b:k:f:nusyvh" OPTION
 do
     case $OPTION in
         V) VERSION_ID="$OPTARG"; VERSION_SPECIFIED=1 ;;
@@ -416,6 +418,7 @@ do
         k) KEYFILE="$OPTARG" ;;
         f) IMAGE_FILE="$OPTARG" ;;
         n) COPY_NET=1;;
+        u) CREATE_UEFI=1;;
         s) INSTALL_TO_SMALLEST=1 ;;
         y) DRY_RUN=1 ;;
         v) set -x ;;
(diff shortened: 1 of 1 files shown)


bin/flatcar-install
```diff
@@ -669,6 +671,15 @@ function write_ignition() if [[ -n "${IGNITION}" ]]; then
     cp "${IGNITION}" "${WORKDIR}/oemfs/config.ign"
 fi
 
+function create_uefi() {
+    ensure_tool "efibootmgr"
+    local EFI_APP="bootx64.efi"
+    if [[ "${BOARD}" = "arm64-usr" ]]; then
+        EFI_APP="bootaa64.efi"
+    fi
+    efibootmgr -c -d ${DEVICE} -l "\efi\boot\${EFI_APP}" -L "Flatcar Container Linux ${CHANNEL_ID}"
```
```suggestion
    efibootmgr -c -d "${DEVICE}" -l "\efi\boot\${EFI_APP}" -L "Flatcar Container Linux ${CHANNEL_ID}"
```
Maybe not needed but still good to quote for consistency

bin/flatcar-install
```diff
@@ -416,6 +417,7 @@ do
         k) KEYFILE="$OPTARG" ;;
         f) IMAGE_FILE="$OPTARG" ;;
         n) COPY_NET=1;;
+        u) CREATE_UEFI=1;;
```
pothos, line 421: Please define it in the block around line 398: `CREATE_UEFI=`

pothos:
Thanks for the PR!

Regarding:
> [ ] flatcar-install: Add an option to create the UEFI flatcar boot entry

The changelog entry will be added in the `changelog/changes/` folder in the coreos-overlay repo when bumping the commit ID in `coreos-base/coreos-init/coreos-init-9999.ebuild`. Can you create this PR, too? In the beginning you can use your branch's commit ID but the final one will be the ID of the merge commit when this PR here gets merged.

bin/flatcar-install
pothos: Ah, and please add it to `print_settings`, too (Edit: Or not? COPY_NET isn't there either…)

jbguerraz: np. but for my information, I wonder why, what differs from `INSTALL_TO_SMALLEST=1` or `COPY_NET=1`, what's the reasoning ?

pothos: Yeah, then let's not add it to `print_settings`

pothos: Oh, the other variables are getting used undefined, I think it's not the best to write a shell script that way, so I suggested to define it first. If there are other variables not following this, no need to care about them now but eventually I hope this script would use `set -u`.

bin/flatcar-install
```diff
@@ -105,6 +105,7 @@ Options:
     -k KEYFILE  Override default GPG key for verifying image signature.
     -f IMAGE    Install unverified local image file to disk instead of fetching.
     -n          Copy generated network units to the root partition.
+    -u          Create UEFI boot entry
```
pothos, line 108:
```suggestion
    -u          Create UEFI boot entry (requires efibootmgr, not available on Flatcar)
```
Maybe we need to document that it won't work when used on Flatcar (e.g., via PXE)? Or do you plan to add the efibootmgr binary to Flatcar?

jbguerraz: Adding it there https://github.com/flatcar-linux/coreos-overlay/pull/1955

jbguerraz:
> Thanks for the PR!
> 
> Regarding:
> 
> > [ ] flatcar-install: Add an option to create the UEFI flatcar boot entry
> 
> The changelog entry will be added in the `changelog/changes/` folder in the coreos-overlay repo when bumping the commit ID in `coreos-base/coreos-init/coreos-init-9999.ebuild`. Can you create this PR, too? In the beginning you can use your branch's commit ID but the final one will be the ID of the merge commit when this PR here gets merged.

Bump PR https://github.com/flatcar-linux/coreos-overlay/pull/1956

pothos (approved):

jepio (approved):

pothos:
Tested it with qemu and the entry got created as expected and after the reboot the second disk got used instead of the primary one

PR #75, opened 2022-06-30
scripts/motdgen: Remove outdated LTS motd information

pothos:
The LTS is now a regular public channel and does not need special docs
for credential configuration. The only thing left that needs attention
is the possibility to freeze the LTS stream to opt-out of automatic
major version updates.

## How to use

Backport to all channels, which may need to use a backport branch in this repo: `flatcar-3033-backport`, `flatcar-3139` and the ones for 3227 needs to be created. ← no, just do it for 3033 and maybe 2605

## Testing done

none

↑ todo in coreos-overlay

### DIFF
--- a/scripts/motdgen
+++ b/scripts/motdgen
@@ -27,14 +27,8 @@ if ! ( mount | grep -q cgroup2 ); then
 	echo -e "Read more: https://kinvolk.io/docs/flatcar-container-linux/latest/container-runtimes/switching-to-unified-cgroups" >> /run/flatcar/motd
 fi
 
-if [[ "${GROUP}" =~ "lts" ]]; then
-	echo "FCL LTS documentation: https://kinvolk.io/fcl-lts-docs" >> /run/flatcar/motd
-fi
-if [[ "${GROUP}" =~ "lts" ]] && { [[ "${DOWNLOAD_USER}" = "" ]] || [[ "${DOWNLOAD_PASSWORD}" = "" ]] ; }; then
-	echo "/etc/flatcar/update.conf: DOWNLOAD_USER and DOWNLOAD_PASSWORD are missing to download LTS updates." >> /run/flatcar/motd
-	echo "Documentation for update.conf: https://kinvolk.io/docs/flatcar-container-linux/latest/clusters/management/update-conf" >> /run/flatcar/motd
-fi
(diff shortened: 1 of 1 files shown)


iaguis (approved):
lgtm

scripts/motdgen
```diff
@@ -27,14 +27,9 @@ if ! ( mount | grep -q cgroup2 ); then
 	echo -e "Read more: https://kinvolk.io/docs/flatcar-container-linux/latest/container-runtimes/switching-to-unified-cgroups" >> /run/flatcar/motd
 fi
 
-if [[ "${GROUP}" =~ "lts" ]]; then
-	echo "FCL LTS documentation: https://kinvolk.io/fcl-lts-docs" >> /run/flatcar/motd
-fi
-if [[ "${GROUP}" =~ "lts" ]] && { [[ "${DOWNLOAD_USER}" = "" ]] || [[ "${DOWNLOAD_PASSWORD}" = "" ]] ; }; then
-	echo "/etc/flatcar/update.conf: DOWNLOAD_USER and DOWNLOAD_PASSWORD are missing to download LTS updates." >> /run/flatcar/motd
-	echo "Documentation for update.conf: https://kinvolk.io/docs/flatcar-container-linux/latest/clusters/management/update-conf" >> /run/flatcar/motd
 fi
```
iaguis: Actually, we should remove this, right?

iaguis (changes requested):

scripts/motdgen
pothos: The variables in https://www.flatcar.org/docs/latest/clusters/management/update-conf? They are still supported by update engine, and are all listed for reference like the other variables even if we don't use them by default such as `PCR_POLICY_SERVER`.

iaguis: I mean the `fi`

iaguis: I mean the `fi`

pothos: Wait, you meant the `fi` ^^

iaguis: I see you've done it already. Thanks!

iaguis (approved):

krnowak:
I'd say that the backports should only be done for flatcar-3033 and flatcar-lts-2605. Other branches are not affected as they never executed the code you just removed.

pothos:
Thanks, yes (and we won't use them to branch a future LTS off)

PR #76, opened 2022-07-27
systemd/system: enable cloudinit for openstack

### LINKED ISSUE #817: Cannot SSH into Flatcar 3227.2.0 instance created in OpenStack
## Description

Cannot SSH into an OpenStack VM created from a Flatcar 3227.2.0 image if the VM is provisioned with an ssh key (not with ignition user_data). Most likely the ssh key is not actually provisioned.

## Impact

Our pipelines that publish Flatcar images to Glance include a basic test that it would be possible to ssh into a VM built from the image and that test fails now.

1. **Set-up**:
  a. Use a Flatcar image downloaded from https://stable.release.flatcar-linux.net/amd64-usr/current/flatcar_production_openstack_image.img.bz2
  b. Create a glance image from it
  c. Create a compute instance based on the image with a key name (e.g. `openstack server create` with `--key-name xxx`, but without `--user-data`)
2. **Task**: Try to ssh into the VM as user `core`
3. **Error**: Public-key authentication is skipped and you get a password prompt

## Expected behavior

ssh with public-key authentication works

## Additional information

I tried to provision the same public key via ignition user-data instead and then ssh worked. So it's probably not a problem with the type of key or host key algorithm. With previous Flatcar image releases provisioning the ssh key worked fine.

Command line used to create a VM (where security group ssh is one that allows incoming port 22)
```
openstack server create <name> --key-name <key-name> --image <image-id> --flavor <flavor-name> --network <network-name> --security-group default --security-group ssh
```

or with terraform, including the key pair creation:
```
resource "tls_private_key" "ssh" {
  algorithm   = "RSA"
  rsa_bits    = 4096
}

resource "openstack_compute_keypair_v2" "ssh" {
  name       = "image_test"
  public_key = tls_private_key.ssh.public_key_openssh
}

resource "openstack_compute_instance_v2" "server" {
  name        = "image_test"

  image_id    = "31115d3f-50d9-4290-9cea-d2c362560290"
  flavor_name = "m1c.tiny"
  key_pair    = openstack_compute_keypair_v2.ssh.name

  security_groups = [
    "default",
    openstack_compute_secgroup_v2.sg_ssh.name,
  ]

  network {
    name = openstack_networking_network_v2.network.name
  }

  depends_on = [
    openstack_networking_subnet_v2.subnet,
  ]
}
```

I tried to understand the difference of the current release to previous ones and my idea is that cloudinit is not run anymore.
Cloudinit used to query the metadata agent via `http://169.254.169.254/2009-04-04/meta-data/public-keys/0/openssh-key` and provide the returned key for the `core` user.

If I configure the VM via user-data I can see in the journal that enabling cloudinit is skipped ("enable-oem-cloudinit.service: Skipped due to 'exec-condition'.") and that is probably because there is already user-data. I couldn't read the journal of a VM that was not configured with user-data though, so I try to guess now, what happened.

In enable-oem-cloudinit.service the ExecConditions should match (.userConfigProvided is probably false if the VM was not configured with user-data). So oem-cloudinit.service will most likely be enabled. But in oem-cloudinit.service there is an ExecCondition that checks if the OEM_ID is one from the list (aws gcp rackspace-onmetal azure cloudsigma packet vmware digitalocean) and openstack is not in it. So cloudinit is not run. But we would need it to fetch the public key.

In older releases there was a default ignition file that created and enabled some oem-cloudinit.service that didn't have that ExecCondition, so cloudinit was run (maybe again only if user-data wasn't available).

bpetermannS11:
# Allow oem-cloudinit.service for openstack too

Fixes https://github.com/flatcar-linux/Flatcar/issues/817, an issue where ssh keys were not fetched if no user-data
was provisioned, but only a keypair.

## How to use

This patch should make sure that a Flatcar-based OpenStack VM can be connected to with ssh if the public key is configured via the key_name server property and not as part of ignite user-data.

## Testing done

none

### DIFF
--- a/systemd/system/oem-cloudinit.service
+++ b/systemd/system/oem-cloudinit.service
@@ -4,8 +4,8 @@ Description=Run cloudinit
 [Service]
 EnvironmentFile=/var/run/ignition.env
 Type=oneshot
-ExecCondition=/usr/bin/bash -xc 'OEMS=(aws gcp rackspace-onmetal azure cloudsigma packet vmware digitalocean); echo $${OEMS[*]} | tr " " "\n" | grep -q -x -F "${OEM_ID}"'
-ExecStart=/usr/bin/bash -xc '/usr/bin/coreos-cloudinit --oem="$(if [ "${OEM_ID}" = aws ]; then echo ec2-compat; elif [ "${OEM_ID}" = gcp ]; then echo gce; else echo "${OEM_ID}" ; fi)"'
+ExecCondition=/usr/bin/bash -xc 'OEMS=(aws gcp rackspace-onmetal azure cloudsigma packet vmware digitalocean openstack); echo $${OEMS[*]} | tr " " "\n" | grep -q -x -F "${OEM_ID}"'
+ExecStart=/usr/bin/bash -xc '/usr/bin/coreos-cloudinit --oem="$(if [ "${OEM_ID}" = aws -o "${OEM_ID}" = openstack ]; then echo ec2-compat; elif [ "${OEM_ID}" = gcp ]; then echo gce; else echo "${OEM_ID}" ; fi)"'
 
 [Install]
 WantedBy=multi-user.target


systemd/system/oem-cloudinit.service
```diff
@@ -4,8 +4,8 @@ Description=Run cloudinit
 [Service]
 EnvironmentFile=/var/run/ignition.env
 Type=oneshot
-ExecCondition=/usr/bin/bash -xc 'OEMS=(aws gcp rackspace-onmetal azure cloudsigma packet vmware digitalocean); echo $${OEMS[*]} | tr " " "\n" | grep -q -x -F "${OEM_ID}"'
-ExecStart=/usr/bin/bash -xc '/usr/bin/coreos-cloudinit --oem="$(if [ "${OEM_ID}" = aws ]; then echo ec2-compat; elif [ "${OEM_ID}" = gcp ]; then echo gce; else echo "${OEM_ID}" ; fi)"'
+ExecCondition=/usr/bin/bash -xc 'OEMS=(aws gcp rackspace-onmetal azure cloudsigma packet vmware digitalocean openstack); echo $${OEMS[*]} | tr " " "\n" | grep -q -x -F "${OEM_ID}"'
+ExecStart=/usr/bin/bash -xc '/usr/bin/coreos-cloudinit --oem="$(if [ "${OEM_ID}" = aws -o "${OEM_ID} = openstack ]; then echo ec2-compat; elif [ "${OEM_ID}" = gcp ]; then echo gce; else echo "${OEM_ID}" ; fi)"'
```
tormath1:
```suggestion
ExecStart=/usr/bin/bash -xc '/usr/bin/coreos-cloudinit --oem="$(if [ "${OEM_ID}" = aws -o "${OEM_ID}" = openstack ]; then echo ec2-compat; elif [ "${OEM_ID}" = gcp ]; then echo gce; else echo "${OEM_ID}" ; fi)"'
```

Thanks for fixing the issue! One thing before running the CI would be to add this commit into: `coreos-init` ebuild file in order to pull the correct version (like I did here: https://github.com/flatcar-linux/coreos-overlay/pull/2061/files) and to open the PR.

bpetermannS11: Added the missing "

tormath1:
Thanks for your contribution - I think we should really start to test again things against openstack.

PR #77, opened 2022-07-27
network/default: exclude `wireguard` type

tormath1:

### DIFF
--- a/systemd/network/zz-default.network
+++ b/systemd/network/zz-default.network
@@ -5,7 +5,7 @@ IPv6AcceptRA=true
 
 [Match]
 Name=*
-Type=!loopback !dummy !bridge !tunnel !vxlan
+Type=!loopback !dummy !bridge !tunnel !vxlan !wireguard
 Driver=!veth
 
 [DHCP]


dongsupark:
I am not completely sure if this change works well for all cases.
So I would say, let's try to run CI tests for all cloud providers.

dongsupark:
Also when looking into the previous PR https://github.com/flatcar-linux/init/pull/30, should we probably fix other places like `yy-pxe.network` and `yy-vmware.network`?

tormath1:
locally tested with:
```yml
---
variant: flatcar
version: 1.0.0
storage:
  files:
    - path: /etc/systemd/network/zz-default.network.d/wg.conf
      contents:
        inline: |
          [Match]
          Type=!wireguard
    - path: /etc/wireguard/kv0.conf
      contents:
        inline: |
          [Interface]
          Address = ...
          PrivateKey = ...

          [Peer]
          PublicKey = ...
          AllowedIPs = ...
          Endpoint = ...
          PersistentKeepalive = 15
systemd:
  units:
    - name: wg-quick@kv0.service
      enabled: true
```

dongsupark (approved):

tormath1:
cherr-picked to:
* flatcar-3227-backport

PR #86, opened 2023-01-13
Optionally check for presence of pv and use - feedback sought

nhi-vanye:
# optionally use PV to provide feedback while writing to disk

If we find pv is installed on the system running flatcar-install, use it to report writing imagefile to disk

## How to use

Install from an image file (not a URL)

This is for feedback - I've only fixed the path I'm using (install from file), happy to extend this to the download path  if requested.

Screenshot 2023-01-13 at 10 47 56 am

### DIFF
--- a/bin/flatcar-install
+++ b/bin/flatcar-install
@@ -10,6 +10,11 @@ function ensure_tool() {
     command -v "${tool}" >/dev/null || { echo "error: command '${tool}' not found" >&2 ; exit 1; }
 }
 
+function have_tool() {
+    local tool="${1}"
+    command -v "${tool}"
+}
+
 if command -v lbzip2 >/dev/null; then
     BZIP_UTIL=lbzip2
 else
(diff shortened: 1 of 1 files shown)


bin/flatcar-install
```diff
@@ -572,9 +586,9 @@ function install_from_file() {
 
     echo "Writing ${IMAGE_FILE}..."
     if [[ "${IMAGE_FILE}" =~ \.bz2$ ]]; then
-        ${BZIP_UTIL} -cd "${IMAGE_FILE}" | write_to_disk
+        ${have_pv:-cat} "${IMAGE_FILE}" | ${BZIP_UTIL} -cd | write_to_disk
```
pothos, line 589: Thanks for your PR!
Could we do `| "${have_pv:-cat}" | write_do_disk` to show the same stats in both cases?

pothos:
Maybe we should also limit the use of `pv` to interactive terminals? Otherwise it's spamming the log output, I guess?

pothos:
Or make it opt-in via an extra flag?

bin/flatcar-install
nhi-vanye: it seems `pv` needs to be the first stage in the pipeline in order to include an ETA; when its streaming between stdin and out it doesn't know the total size...

Fitting it into the other places would obviously be easier in stream mode...

So I guess its code simplicity (streaming) vs features (first in pipeline) - I'm not tied to having an ETA so do you have a preferred approach ?

nhi-vanye:
> Maybe we should also limit the use of `pv` to interactive terminals? Otherwise it's spamming the log output, I guess?

Doh! Yes will do.

PR #88, opened 2023-01-31
If the directory symlink exists, don't create a symlink in it

pothos:
When motdgen ran again, we ended up with a "flatcar -> flatcar" symlink inside the /run/flatcar/ directory because the existing /run/coreos symlink is pointing to the /run/flatcar directory and without -T ln will then create the symlink inside the given path instead of overwriting the symlink.
Add the -T flag and also use it in the Makefile whenenver we want to set up symlinks for directories.

## How to use

It's not a bug that makes problems, no backport needed.

## Testing done

Ran the new `motdgen` version here with `sudo` and it didn't create the wrong `/run/flatcar/flatcar` symlink anymore while `sudo /usr/lib/flatcar/motdgen` did.


krnowak (approved):

pothos:
Backporting to Beta because along with another change

PR #89, opened 2023-02-16
fix(systemd/network): exclude node* and kube* from dhcp by default

maximee-leroyy:
# [systemd/network: exclude node* and kube* from dhcp by default]

Currently, when using `kubeadm join`, it says that the command works but sometimes it fails because the container `nodelocaldns` creates a dummy interface called `nodelocaldns` by default and `systemd-networkd` manages it with the default file `zz-default.network` that adds DHCP and IPV6RA support.

So when `nodelocaldns` tries to do an equivalent of `ip addr add`, the command works but it will silently fail because the interface is managed by two different programs (`systemd-networkd` and `nodelocaldns`)

## How to use

Check that the interfaces created by `kube-proxy` and `nodelocaldns` are not managed by `systemd-networkd` with the command `networkctl`.

Check that `kubeadm join` properly works so the node can join the cluster by setting its default `nodelocaldns` interface to the IPs defined with `-localip` in the `node-cache` container.

## Testing done

Checked that everything in `How to use` works

### DIFF
--- a/systemd/network/kubedns.network
+++ b/systemd/network/kubedns.network
@@ -1,0 +1,9 @@
+# Exclude Kubernetes local node dns dummy interfaces from DHCP by default
+# https://github.com/kubernetes/dns/blob/da9249d88ca22ecdf09e25c98f3c0c572a85d34b/cmd/node-cache/main.go#L83
+
+[Match]
+Name=nodelocaldns
+Driver=dummy
+
+[Link]
+Unmanaged=yes
--- a/systemd/network/kubeproxy.network
+++ b/systemd/network/kubeproxy.network
@@ -1,0 +1,9 @@
+# Exclude Kubernetes proxy dummy interfaces from DHCP by default
+# https://github.com/kubernetes/kubernetes/blob/cd5ab497413faacae4e712b487c63c8b45c292f5/pkg/proxy/ipvs/proxier.go#L91
+
+[Match]
+Name=kube-ipvs0
+Driver=dummy
+
+[Link]
+Unmanaged=yes


maximee-leroyy:
cc @pothos @krnowak

systemd/network/kubedns.network
```diff
@@ -0,0 +1,7 @@
+# Exclude Kubernetes local node dns dummy interfaces from DHCP by default
+
+[Match]
+Name=node*
```
pothos: That's quite generic, can't we make it more specific? E.g., if there are more names with `node…` we can use a list of the longest prefixes. Also, we can match against the driver that is used.
Ideally kubeadm would create this rule here itself under `/run/systemd/network/` and match by the MAC address or exact interface name…
Can you add a docs/source link in the commit message or even here in the file as comment?

systemd/network/kubeproxy.network
```diff
@@ -0,0 +1,7 @@
+# Exclude Kubernetes proxy dummy interfaces from DHCP by default
+
+[Match]
+Name=kube*
```
pothos: It would also be good to be more specific here and also point at docs for this interface/where it comes from, and maybe make an upstream issue to itself create the network files under `/run/systemd/network/`.

systemd/network/kubedns.network
pothos: There is `Path=`, `Driver=`, `Type=`, `Kind=` and more in https://www.freedesktop.org/software/systemd/man/systemd.network.html#%5BMatch%5D%20Section%20Options with `networkctl` commands how to show the current values.

maximee-leroyy: I added a link as a comment that points to the code where this name originates.
Unfortunately it is configurable with parameters so its not a fixed value but at least we can handle only the default name if you prefer

systemd/network/kubeproxy.network
maximee-leroyy: I added a link as a comment that points to the code where this name originates.
I can only be more precise with the name

systemd/network/kubedns.network
pothos: Thanks, that already helps. If you have a cluster around to run `networkctl status INTERFACE` it would be great to know if we can further make this more specific with the type=, kind=, and driver= settings.

pothos: (Next thing I'll do then is to start a build with this PR that runs through all cloud provider k8s tests)

maximee-leroyy:
```
kube-ipvs0                                                         
                     Link File: /usr/lib/systemd/network/99-default.link
                  Network File: n/a
                          Type: ether
                         State: routable (unmanaged)
                  Online state: unknown
                        Driver: dummy
                           MTU: 1500
                         QDisc: noqueue
  IPv6 Address Generation Mode: eui64
          Queue Length (Tx/Rx): 1/1
             Activation Policy: up
           Required For Online: yes
```

maximee-leroyy:
```
nodelocaldns                                                      
                     Link File: /usr/lib/systemd/network/99-default.link
                  Network File: n/a
                          Type: ether
                         State: off (unmanaged)
                  Online state: unknown
                        Driver: dummy
                           MTU: 1500
                         QDisc: noop
  IPv6 Address Generation Mode: eui64
          Queue Length (Tx/Rx): 1/1
             Activation Policy: up
           Required For Online: yes
```

pothos: Thanks! That means we can add `Driver=dummy` for both and it won't match any physical or veth container interfaces.

systemd/network/kubedns.network
```diff
@@ -0,0 +1,8 @@
+# Exclude Kubernetes local node dns dummy interfaces from DHCP by default
+# https://github.com/kubernetes/dns/blob/da9249d88ca22ecdf09e25c98f3c0c572a85d34b/cmd/node-cache/main.go#L83
+
+[Match]
+Name=nodelocaldns
+
```
pothos, line 7:
```suggestion
Driver=dummy

```

systemd/network/kubeproxy.network
```diff
@@ -0,0 +1,8 @@
+# Exclude Kubernetes proxy dummy interfaces from DHCP by default
+# https://github.com/kubernetes/kubernetes/blob/cd5ab497413faacae4e712b487c63c8b45c292f5/pkg/proxy/ipvs/proxier.go#L91
+
+[Match]
+Name=kube-ipvs0
+
```
pothos, line 7:
```suggestion
Driver=dummy

```

pothos (approved):

pothos:
Thanks

pothos:
I actually had excluded `dummy` devices from being managed by default. Have to check why this is not working: https://github.com/flatcar/init/blob/flatcar-master/systemd/network/zz-default.network#L8

pothos:
Ah, `Type` vs `Driver`!

pothos:
Fix prepared in https://github.com/flatcar/init/pull/90

We can still keep your changes, we also have this for other similar cases and these are probably the ones we would backport. I still would love to see Kubernetes to provide the unit itself in `/run/systemd/network/` instead of forcing distros and users to take care of this.

PR #90, opened 2023-02-20
systemd/network: Fix dummy interface exclusion rule in default networkd unit

pothos:
The default networkd unit should not apply to all devices, e.g., not to bridges, tunnels, dummy interfaces etc. However, for dummy interfaces the rule was wrong.
Fix the exclusion rule for dummy interfaces, which are not a type but a certain driver.

## Testing done

Was tested by copying it to `/etc/systemd/network/` and doing `ip link add dummy0 type dummy` to create such an interface and `networkctl reload` to apply the unit change and see that the interface becomes unmanaged.

### DIFF
--- a/systemd/network/zz-default.network
+++ b/systemd/network/zz-default.network
@@ -5,8 +5,8 @@ IPv6AcceptRA=true
 
 [Match]
 Name=*
-Type=!loopback dummy bridge tunnel vxlan wireguard
-Driver=!veth
+Type=!loopback bridge tunnel vxlan wireguard
+Driver=!veth dummy
 
 [DHCP]
 UseMTU=true


krnowak:
I don't know how to verify this PR. I was looking for a list of possible values for the `Driver` and `Type` fields and the closest thing I found was in [`systemd.netdev(5)`](https://man.archlinux.org/man/systemd.netdev.5.en), but it has a single table that contains values that you can find in both `Type` and `Driver` fields.

pothos:
Observing behavior like in natural science, not following rules on paper like in math or engineering ;)

Actually it would be nice to submit our default network unit file as example in the systemd docs to spare others running into the same problems.

krnowak:
> Observing behavior like in natural science, not following rules on paper like in math or engineering ;)

Yeah, I was being lazy. :) Repeated the steps in a VM, seems to be doing the trick.

> 
> Actually it would be nice to submit our default network unit file as example in the systemd docs to spare others running into the same problems.

Right, there seem to be no example for configuring links as unmanaged.

krnowak (approved):

PR #94, opened 2023-03-28
flatcar-reset: Add backup and preview

pothos:
When users run flatcar-reset from the command line to try it out, this should not result in data loss when they forget to quote the keep paths correctly or didn't know how to stop the staged reset. Add a warning and the option to see the list of files in preview. Also add the option to do a backup when staging a reset. The backup does not cover ACLs because Flatcar's rsync isn't compiled with support for them.

## How to use

Improve our rsync to support ACLs even though most people don't use them

## Testing done

Copied it over manually to a VM at ` /home/core/flatcar-reset`

```
touch myfile
sudo ./flatcar-reset --preview-keep --keep-paths /etc/ssh /home/core/flatcar-reset
sudo ./flatcar-reset --preview-delete --keep-paths /var/log /home/core/flatcar-reset
sudo ./flatcar-reset --backup --keep-paths /home/core/flatcar-reset # takes some time because of the checks to correctly preserve the directory structure
sudo reboot
ls myfile # does not exist
sudo ./flatcar-reset --restore-backup
ls myfile # exists again
```

### DIFF
--- a/bin/flatcar-reset
+++ b/bin/flatcar-reset
@@ -2,14 +2,20 @@
 set -euo pipefail
 
 # The regex path arguments for --keep-paths are treated as regular arguments
-opts=$(getopt --name "$(basename "${0}")" --options 'hF:U:KM' \
-       --longoptions 'help,ignition-file:,ignition-url:,keep-paths,keep-machine-id' -- "${@}")
+opts=$(getopt --name "$(basename "${0}")" --options 'hF:U:KMBSPWDR' \
+       --longoptions 'help,ignition-file:,ignition-url:,keep-paths,keep-machine-id,backup,stop,preview-delete,preview-keep,delete-backup,restore-backup' -- "${@}")
 eval set -- "${opts}"
 
 KEEPMACHINEID=
 IGNITIONFILE=
 IGNITIONURL=
 HASKEEPPATHS=
+BACKUP=
+STOP=
+PREVIEWDELETE=
+PREVIEWKEEP=
+DELETEBACKUP=
+RESTOREBACKUP=
 
 while true; do
   case "$1" in
@@ -25,20 +31,25 @@ while true; do
     echo "  preserved as well because MYPATH/.* is automatically appended as additonal regular expression for paths to keep."
     echo "  To delete the contents of a folder but keep the folder itself, specify it as equivalent regular expression in the form of"
     echo "  '^/etc/mypath', '/etc/mypath$', '/etc/mypat[h]', '/etc/(mypath)', or '(/etc/mypath)'. The used regular expression language"
-    echo "  is that of egrep. Assuming you specified '/etc/mypath', you can test which paths will be deleted with (note the '-not'):"
-    echo "    find / /etc -xdev -regextype egrep -not -regex '(/etc/mypath|/etc/mypath/.*)'"
-    echo "  You can tests which path will be kept with with (note the absence of '-not'):"
-    echo "    find / /etc -xdev -regextype egrep -regex '(/etc/mypath|/etc/mypath/.*)'"
-    echo "  Both / and /etc have to be specified because /etc is an overlay mount."
+    echo "  is that of egrep."
     echo "  Meaningful examples are:"
     echo "  - '/etc/ssh/ssh_host_.*' to preserve SSH host keys"
     echo "  - '/var/log' to preserve system logs"
     echo "  - '/var/lib/docker' '/var/lib/containerd' to preserve container state and images"
+    echo "  The rootfs does not include the files from /etc that are provided by the overlay mount unless they were copied up."
+    echo "  Therefore, you won't see them in the preview or backup."
     echo "Options:"
     echo "  -F, --ignition-file <FILE>	Writes the given Ignition config JSON file to /usr/share/oem/config.ign"
     echo "  -U, --ignition-url <URL>	Writes the given Ignition config JSON URL as kernel cmdline parameter to /usr/share/oem/grub.cfg"
     echo "  -K, --keep-paths <REGEX>...	Writes the given regular expressions for paths to keep as combined OS reset info to /selective-os-reset"
-    echo "  -M, --keep-machine-id	Writes the current machine ID as kernel cmdline parameter to /usr/share/oem/grub.cfg to preserve it"
+    echo "  -M, --keep-machine-id		Writes the current machine ID as kernel cmdline parameter to /usr/share/oem/grub.cfg to preserve it"
+    echo "  -B, --backup			Copies the files that will be deleted to /flatcar-backup/"
+    echo "Actions (exclusive, no OS reset will be staged):"
+    echo "  -S, --stop			Stops the staged OS reset and nothing will happen on the next boot"
+    echo "  -P, --preview-delete		Prints the files that will be deleted"
+    echo "  -W, --preview-keep		Prints the files that will be kept"
+    echo "  -D, --delete-backup		Deletes the backup under /flatcar-backup/"
+    echo "  -R, --restore-backup		Copies the files from /flatcar-backup/ to their original location on the rootfs"
     echo
     echo "Example for selectively resetting the OS with retriggering Ignition while keeping SSH host keys, logs, and machine ID:"
     echo "  sudo $(basename "${0}") --keep-machine-id --keep-paths '/etc/ssh/ssh_host_.*' /var/log"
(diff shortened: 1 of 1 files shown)


bin/flatcar-reset
```diff
@@ -89,8 +118,110 @@ if [ "${HASKEEPPATHS}" = 1 ]; then
   done
 fi
```
krnowak, line 120: I'd add a check if really only one exclusive action was specified. Maybe something like:
```bash
teststr="${STOP}${PREVIEWDELETE}${PREVIEWKEEP}${DELETEBACKUP}${RESTOREBACKUP}"
if [[ -n "${teststr}" ]] && [[ "${teststr}" != '1' ]]; then echo 'one exclusive action allowed'; exit 1; fi

File: bin/flatcar-reset [hunk 2]
```diff
@@ -89,8 +118,110 @@ if [ "${HASKEEPPATHS}" = 1 ]; then
...
+# Handle exclusive actions
+
+if [ "${STOP}" = 1 ]; then
+  echo "Removing /selective-os-reset and /boot/flatcar/first_boot"
+  rm -f "/selective-os-reset" "/boot/flatcar/first_boot"
+  exit 0
+elif [ "${PREVIEWDELETE}" = 1 ]; then
+  # For find -not means that we look at all files that are not matched by the keep regex
+  walkroot echo -not
+  exit 0
```
krnowak: The output of --preview-delete is kinda scary at first, because it tells you that `/bin` and `/lib` and `/lib64` will be dropped. Maybe instead of `echo` there should be an action that prints the entry and if the entry is a symlink, it could print its destination (so something like `/usr (symlink to usr/bin)`).

krnowak: Also, prints `/.etc-work` and `/.etc-work/work`. Maybe there ought to be some explanation description of the two?

bin/flatcar-reset
```diff
@@ -89,8 +118,110 @@ if [ "${HASKEEPPATHS}" = 1 ]; then
...
+    local tocreate=""
+    while IFS= read -r -d '' pathpart; do
+      if [ "${pathpart}" = "" ]; then
+        continue
+      fi
+      tocreate="${tocreate}/${pathpart}"
+      if [ ! -d "/flatcar-backup${tocreate}" ]; then
+        # Use rsync to create the directory with the right permissions but don't copy its contents
+        # TODO: --acls
+        unshare -m sh -c "umount /etc && rsync -x -a --exclude='*' '${tocreate}' '/flatcar-backup${tocreate}'"
```
rata, line 213: Maybe we also want: `--hard-links, --sparse, --delete`?

I'm unsure if the hard-links option will behave as expected, though. You will have to test that.

The sparse flag seems nice, in case there are sparse files so they don't take lot of space in the backup (more than they took before backup).

And the delete flag I think it is needed to delete things from previous backups. I _think_ (haven't tested) this will just add files to the backup path, and never delete them. With the --delete rsync flag we should only keep the files from the last time the backup was run.

Am I missing something?

bin/flatcar-reset
```diff
@@ -89,8 +118,110 @@ if [ "${HASKEEPPATHS}" = 1 ]; then
...
+      fi
+    done < <(dirname -z "${entry}" | tr '/' '\0')
+  fi
+  # TODO: --acls
+  unshare -m sh -c "umount /etc && if [ -d '${entry}' ] && [ ! -L '${entry}' ]; then rsync -x -a --exclude='*' '${entry}/' '${newpath}/'; else cp -a '${entry}' '${newpath}'; fi"
```
rata, line 218: Idem here?

bin/flatcar-reset
rata: Hmm, it seems we a `rm -rf` before, which seems better than the delete flags. So, IIUC, the delete flag is not needed and ignore that part :D

bin/flatcar-reset
```diff
@@ -89,8 +118,110 @@ if [ "${HASKEEPPATHS}" = 1 ]; then
...
+  rm -rf "/flatcar-backup/"
+  exit 0
+elif [ "${RESTOREBACKUP}" = 1 ]; then
+  if [ ! -d "/flatcar-backup/" ]; then
+    echo "Error: The directory /flatcar-backup/ does not exist" > /dev/stderr
+    exit 1
+  fi
+  echo "Restoring rootfs files from /flatcar-backup/"
+  # TODO: our rsync does not support --acls
+  unshare -m sh -c "umount /etc && rsync -x -a --inplace -v /flatcar-backup/ /"
```
rata: Just curious, if you don't mind me asking: how is the etc overlay mount configured? I mean, what is the lower/upper dir?

Out of curiosity, I wanted to understand what we are really doing here: we create a new mount namespace, unmount etc and then write this folder. But in the host mount namespace, that folder is used as the upper/lower dir in the /etc mount, right? Doesn't the kernel dmesg show a log line saying we are modifying the lower/upper dir behind its back anyways in the host mount namespace, even if we create a new mount namespace  and all?

rata (commented):
Thanks A LOT for this PR. I really think this is super useful.

bin/flatcar-reset
pothos: Thanks

bin/flatcar-reset
pothos: How about a message at the end like `Note that it is ok to delete the /bin or /lib symlinks and any other OS files/directories like /etc or /.etc-work as they will be recreated.`

bin/flatcar-reset
pothos: Thanks, I think hard links would violate what I expect from a backup: it should not change after creation, but with hardlinks the files would still be written to during the usage until reboot and during the reboot.

The rsync call is just to create a direcotry, the actual files get copied with `cp -a`. I think `--sparse=auto` and `--reflink=auto` is the default behavior? Not sure I understand what `cp --help` says.

bin/flatcar-reset
pothos: The lowerdir is `/usr/share/flatcar/etc`, the upperdir is `/etc`. Yes, the modification on the fly is not really supported by overlayfs. We could remount /etc (maybe with `umount -l /etc` first) or something like that to set the overlayfs up again?

pothos: At least a reboot should be recommended afterwards, will add that.

bin/flatcar-reset
krnowak: Sounds good.

bin/flatcar-reset
rata: Oh, makes sense.

In my experience long ago, when I used sparse files for VMs locally, I had to add those flags to cp. I don't know if the heuristics got better. In any case, I think the flags won't hurt and also not being there won't probably hurt much as it is uncommon for etc files to be sparse.

bin/flatcar-reset
rata: Yeah, that seems nice. Maybe another option is to not change the content now, but add some sentinel or something that on next boot, the restore is done before mounting or the mounting is done using another lower path.

Another idea from the top of my head: let's say you have `etc-a` and `etc-b` dirs, and a symlink `current` pointing to the current dir. Then you can change the symlink now and on next boot it will mount the old dir that was used.

But I'm not familiar with the flatcar etc setup to suggest something that really fits in the setup.

pothos: Doing it from the initramfs would be the safest way, I'm just afraid of adding more things to the initrd as it is already fairly complex. It could also conflict with Ignition running again, so for now I hope it's good enough to do a reboot or the user could even just copy the wanted files instead of restoring everything.

The OS reset is about the whole rootfs, not just `/etc`, we would also have to do it for `/var` or single files and so on.

bin/flatcar-reset
pothos: The behavior of `--reflink=always --sparse=always` was a bit strange at least in older cp versions where it would create sparse files while it could have rather done a simple reflink for the whole file. I think that was improved even more in the latest version, didn't try again.

bin/flatcar-reset
rata: Makes sense!

bin/flatcar-reset
rata: Cool, so I trust what you think it is better :)

pothos:
I would merge this soon if there are no other comments

bin/flatcar-reset
pothos: In the future the `/etc` overlay could be managed by systemd-confext (former planned name was systemd-syscfg), and the remounting could be triggered safely when it uses https://lwn.net/Articles/927491/

rata: haha, thought the same when I saw the article :)

krnowak (approved):

PR #95, opened 2023-04-04
/usr/share/oem -> /oem

krnowak:
CI: http://jenkins.infra.kinvolk.io:8080/job/container/job/packages_all_arches/1784/cldsv


bin/flatcar-install
```diff
@@ -74,7 +74,7 @@ CHANNEL_ID=${GROUP:-stable}
 BOARD=$(default_board)
 
 OEM_ID=
-for f in /usr/share/oem/oem-release /etc/oem-release; do
+for f in /oem/oem-release /usr/share/oem/oem-release /etc/oem-release; do
```
pothos, line 77: I wonder if people want to run this script on an older Flatcar version, e.g., to have the latest gpg subkey. But then i wonder why they wouldn't update the install OS. Not sure if we really want to support this case but it's easily possible.

pothos: Meaning, thanks for doing it^^

pothos (approved):

sayanchowdhury (approved):

PR #96, opened 2023-04-17
ensure-sysext: drop daemon-reload command

tormath1:
systemd-sysext will perform a daemon-reload for each merge / unmerge / refresh action.

<hr>

Related to: https://github.com/flatcar/Flatcar/issues/984

:warning: this needs a systemd upgrade to bring the reload mechanism from the extension


systemd/system/ensure-sysext.service
```diff
@@ -11,7 +11,6 @@ ConditionDirectoryNotEmpty=|/usr/lib/extensions
 [Service]
 Type=oneshot
 RemainAfterExit=yes
-ExecStart=/usr/bin/systemctl daemon-reload
 ExecStart=/usr/bin/systemctl restart --no-block sockets.target timers.target multi-user.target
```
pothos: Most things are also converted to use `Upholds=`, I think we should also drop this line here (and call it out in the changelog).

tormath1: If we drop this line too we can actually drop this unit.

pothos:
I think we can use this now

tormath1:
> I think we can use this now

Oh yes - 1y later! Rebased and kicked off a CI, let's see how it goes.

tormath1:
Closed in favor of: https://github.com/flatcar/init/pull/138

PR #97, opened 2023-06-14
bin/flatcar-install: run write_to_disk from subshell

tormath1:
Somehow, with the bash upgrade the trap RETURN signal is not handled correctly (it's ignored).

Adding an explicit return makes it work.

<hr>

The bash upgrade and the way it's done (a single huge commit) it's too complicated to make a proper bisect but I assume the issue is regarding the changes on how trap signals are handled.

## Testing done
* Locally
* Implementing Mantle test for this script

### DIFF
--- a/bin/flatcar-install
+++ b/bin/flatcar-install
@@ -704,7 +704,9 @@ function install_from_url() {
     fi 3> >(write_to_disk)
 }
 
-function write_cloudinit() if [[ -n "${CLOUDINIT}${COPY_NET}" ]]; then
+function write_cloudinit() {
+  if [[ -n "${CLOUDINIT}${COPY_NET}" ]]; then
+    (
     # The ROOT partition should be #9 but make no assumptions here!
     # Also don't mount by label directly in case other devices conflict.
     local ROOT_DEV=$(blkid -t "LABEL=ROOT" -o device "${DEVICE}"*)
@@ -714,7 +716,7 @@ function write_cloudinit() if [[ -n "${CLOUDINIT}${COPY_NET}" ]]; then
       "btrfs") mount -t btrfs -o subvol=root "${ROOT_DEV}" "${WORKDIR}/rootfs" ;;
       *)       mount "${ROOT_DEV}" "${WORKDIR}/rootfs" ;;
     esac
-    trap 'umount "${WORKDIR}/rootfs"' RETURN
+    trap 'umount "${WORKDIR}/rootfs"' EXIT
 
     if [[ -n "${CLOUDINIT}" ]]; then
       echo "Installing cloud-config..."
@@ -727,20 +729,26 @@ function write_cloudinit() if [[ -n "${CLOUDINIT}${COPY_NET}" ]]; then
       # Copy the entire directory, do not overwrite anything that might exist there, keep permissions, and copy the resolve.conf link as a file.
       cp --recursive --no-clobber --preserve --dereference /run/systemd/network/* "${WORKDIR}/rootfs/etc/systemd/network"
     fi
-fi
+    )
+  fi
+}
 
-function write_ignition() if [[ -n "${IGNITION}" ]]; then
+function write_ignition() {
+  if [[ -n "${IGNITION}" ]]; then
+    (
     # The OEM partition should be #6 but make no assumptions here!
     # Also don't mount by label directly in case other devices conflict.
     local OEM_DEV=$(blkid -t "LABEL=OEM" -o device "${DEVICE}"*)
 
     mkdir -p "${WORKDIR}/oemfs"
     mount "${OEM_DEV}" "${WORKDIR}/oemfs" || { btrfstune -f -u "${OEM_DEV}" ; mount "${OEM_DEV}" "${WORKDIR}/oemfs" ; }
-    trap 'umount "${WORKDIR}/oemfs"' RETURN
+    trap 'umount "${WORKDIR}/oemfs"' EXIT
 
     echo "Installing Ignition config ${IGNITION}..."
     cp "${IGNITION}" "${WORKDIR}/oemfs/config.ign"
-fi
+    )
+  fi
+}
 
 function create_uefi() {
     ensure_tool "efibootmgr"


bin/flatcar-install
```diff
@@ -596,7 +596,7 @@ function write_to_disk() {
         echo "Failed to reread partitions on ${DEVICE}" >&2
     done
     [ -z "$try" ] || exit 1
-    udevadm settle
+    udevadm settle && return
```
pothos: Would this work, too, because when udevadm settle fails, the script exists due to  `set -e`?
```suggestion
    udevadm settle
    return
```

pothos: A comment for this workaround would also be good

krnowak:
I remember replacing all RETURN traps to EXIT traps in subshells because of that. See it here: https://github.com/flatcar/scripts/commit/eee6b50aa57fae7e6b12f1da3ef7356d335a20c8

tormath1:
> I remember replacing all RETURN traps to EXIT traps in subshells because of that. See it here: [flatcar/scripts@eee6b50](https://github.com/flatcar/scripts/commit/eee6b50aa57fae7e6b12f1da3ef7356d335a20c8)

Wow, that's interesting thanks for sharing. Interesting to to see that we did not get the issue with 5.1 but in 5.2...

bin/flatcar-install
tormath1: Thanks for the suggestion - it works fine.

pothos:
The install script is tested in the Equinix Metal installation path, strange that we didn't see the issue there?

bin/flatcar-install
pothos: Why do we need the trap? Can't the write the "done" in the last line instead of the `return`?

pothos: Ok, I think the idea was to write the "done" as soon as the function did something, instead of the last line it should be done as first.

pothos: The current approach with the explicit `return` doesn't cover the case where `dd` failed and thus the `is_modified` check wouldn't work anymore.

tormath1:
> The install script is tested in the Equinix Metal installation path, strange that we didn't see the issue there?

Thanks, I'm now even more confused. :D

bin/flatcar-install
tormath1: I followed @krnowak approach and use a subshell + EXIT trap instead of explicit `return` - but I still can't explain why it works on Equinix Metal setup.

bin/flatcar-install
```diff
@@ -577,8 +577,11 @@ function wait_for_disk() {
 }
 
 function write_to_disk() {
+    # We run everything from a subshell and we use an EXIT trap
+    # to emulate previous RETURN trap.
+    (
     mkfifo -m 0600 "${WORKDIR}/disk_modified"
-    trap '(exec 2>/dev/null ; echo done > "${WORKDIR}/disk_modified") &' RETURN
+    trap '(exec 2>/dev/null ; echo done > "${WORKDIR}/disk_modified") &' EXIT
```
```suggestion
    (exec 2>/dev/null ; echo done > "${WORKDIR}/disk_modified") &
```
I'm fine with the current approach, my idea was to simplify this and don't use traps.
This should work because we don't wait for the write somewhere in parallel but it's all sequential code and the marker can be written early.

tormath1: Removing the trap is creating a weird behavior:
```
++ write_to_disk
++ mkfifo -m 0600 /tmp/flatcar-install.r3mrgIyVoA/disk_modified
+ wget --tries 10 --timeout=20 --retry-connrefused --no-verbose -O - https://beta.release.flatcar-linux.net/amd64-usr/3602.1.0/flatcar_production_image.bin.bz2
+ gpg --batch --trusted-key E25D9AED0593B34A --verify /tmp/flatcar-install.r3mrgIyVoA/flatcar_production_image.bin.bz2.sig -
+ tee /dev/fd/62
++ lbzip2 -cd
+++ blockdev --getsz /dev/sda
++ exec
++ dd conv=nocreat count=1024 if=/dev/zero of=/dev/sda seek=41942016 status=none
++ dd bs=1M conv=nocreat of=/dev/sda status=none
2023-06-23 09:40:39 URL:https://beta.release.flatcar-linux.net/amd64-usr/3602.1.0/flatcar_production_image.bin.bz2 [379416813/379416813] -> "-" [1]
gpg: Signature made Mon May 29 09:40:52 2023 UTC
gpg:                using RSA key 8D6DA7853CFE1B1ED346ED0DEDBDF411267EC954
gpg:                issuer "buildbot@flatcar-linux.org"
gpg: key E25D9AED0593B34A marked as ultimately trusted
gpg: checking the trustdb
gpg: marginals needed: 3  completes needed: 1  trust model: pgp
gpg: depth: 0  valid:   1  signed:   0  trust: 0-, 0q, 0n, 0m, 0f, 1u
gpg: Good signature from "Flatcar Buildbot (Official Builds) <buildbot@flatcar-linux.org>" [ultimate]
+ wait_for_disk
+ '[' -n '' ']'
+ read -rt 7200 _disk_status
+ write_cloudinit
+ [[ -n '' ]]
+ write_ignition
+ [[ -n '' ]]
+ [[ -n '' ]]
+ echo 'Success! Flatcar Container Linux beta 3602.1.0 is installed on /dev/sda'
Success! Flatcar Container Linux beta 3602.1.0 is installed on /dev/sda
+ rm -rf /tmp/flatcar-install.r3mrgIyVoA
+ trap - EXIT
```

It ends right after the `dd` (it does not go through the `udevadm settle` and the for loop) so it takes a few second before the changes being visible

pothos: Ah, `install_from_url` doesn't wait for `write_to_disk` to finish. It even doesn't capture the error code because  `> >(write_to_disk)` redirects the output to the input FD but it's not a pipe where `pipefail` would work. This should be fixed to properly fail on error.

pothos: Prepared a fix in https://github.com/flatcar/init/pull/99

pothos: Had to touch the waiting logic anyway now to signal the failure and removed the trap.

pothos (commented):
There are two other `RETURN` traps for `umount`, do you want to use the subshell trick there?

tormath1:
@pothos done and tested with a merge of your PR and mine. :heavy_check_mark:

pothos (approved):

PR #98, opened 2023-06-22
Move ssh configs to coreos-base/misc-files package in our overlay

krnowak:
The ssh and sshd configuration in Gentoo has been changed from a single file thing to a single file config including configuration snippets in /etc/ssh{,d}_config.d/ directory. We will ride on this change and turn our ssh configs into such snippets. But these will be installed by a different package directly into /etc/ssh directory. This makes the tmpfiles config file creating symlinks for ssh configs unnecessary, so drop it too.


configs/sshd_config
```diff
@@ -1,17 +0,0 @@
-# Use most defaults for sshd configuration.
```
pothos, line 1: Instead of re-adding this through the misc package, could we change the install directive to put it into `/etc` (but I guess it's a taste question of either having files in a repo like here or under the `files/` ebuild subfolder).

I think we also need one of those compat symlinks for `/usr/share/ssh/ssh(d)_config` to point to the file under `/usr/share/flatcar/etc` or something like that. The old symlink for existing installations could stick around and thus require that target to be present.

krnowak: > Instead of re-adding this through the misc package, could we change the install directive to put it into `/etc` (but I guess it's a taste question of either having files in a repo like here or under the `files/` ebuild subfolder).
> 

Yeah, my opinion was that this repo is no place for ssh configs, Thus the move.

> I think we also need one of those compat symlinks for `/usr/share/ssh/ssh(d)_config` to point to the file under `/usr/share/flatcar/etc` or something like that. The old symlink for existing installations could stick around and thus require that target to be present.

These are done by coreos-base/misc-files in the scripts PR.

krnowak:
This will be pulled into image as a part of https://github.com/flatcar/scripts/pull/945.

pothos (approved):

PR #101, opened 2023-07-28
flatcar-update: Support Flatcar OEM and extension payloads

pothos:
The OEMs are now getting ported over to systemd-sysext images and they are delivered as additional update payloads in the Omaha response. We also define optional Flatcar extensions that the user can enable. While update-engine's post-install action and the initrd have a fallback mechanism that use the release server in case flatcar-update does not provide the required payloads, this does not work for airgapped environments or updating to developer payloads.
Let flatcar-update download the required payloads for the running machine from the release server instead of relying on any fallback logic and also request the user to provide any required extension payloads.

## Testing done

See https://github.com/flatcar/update_engine/pull/24

### DIFF
--- a/bin/flatcar-update
+++ b/bin/flatcar-update
@@ -239,8 +239,12 @@ tee -a /tmp/response > /dev/null <<-EOF
 	</updatecheck><event status="ok"></event></app></response>
 EOF
 
-trap "umount /usr/share/update_engine/update-payload-key.pub.pem 2> /dev/null || true; rm -f /tmp/response /tmp/payload-server ; kill 0" EXIT INT
+true > /tmp/payload-server-pids
+trap "umount /usr/share/update_engine/update-payload-key.pub.pem 2> /dev/null || true; rm -f /tmp/response /tmp/payload-server ; cat /tmp/payload-server-pids | xargs -r kill ; rm -f /tmp/payload-server-pids" EXIT INT
 ncat --keep-open -c "echo -en 'HTTP/1.1 200 OK\ncontent-type: text/xml\ncontent-length: $(stat --printf='%s\n' /tmp/response)\n\n'; cat /tmp/response" -l "$LISTEN_PORT_1" &
+CHILDPID="$!"
+echo "${CHILDPID}" >> /tmp/payload-server-pids
+
 
 # Helper script because inline quoting is insane
 tee /tmp/payload-server > /dev/null <<'EOF'
@@ -267,6 +271,8 @@ EOF
 
 chmod +x /tmp/payload-server
 socat TCP-LISTEN:"${LISTEN_PORT_2}",reuseaddr,fork SYSTEM:'/tmp/payload-server /var/tmp/flatcar-update/ application/gzip' &
+CHILDPID="$!"
+echo "${CHILDPID}" >> /tmp/payload-server-pids
 
 if [ "${FORCE_DEV_KEY}" = "1" ] || [ "${FORCE_FLATCAR_KEY}" = "1" ]; then
   rm -f /tmp/key
--- a/bin/flatcar-update
+++ b/bin/flatcar-update
@@ -1,13 +1,14 @@
 #!/bin/bash
 set -euo pipefail
 
-opts=$(getopt --name "$(basename "${0}")" --options 'hV:P:L:M:DFA' \
-       --longoptions 'help,to-version:,to-payload:,listen-port-1:,listen-port-2:,force-dev-key,force-flatcar-key,disable-afterwards' -- "${@}")
+opts=$(getopt --name "$(basename "${0}")" --options 'hV:P:E:L:M:DFA' \
+       --longoptions 'help,to-version:,to-payload:,extension:,listen-port-1:,listen-port-2:,force-dev-key,force-flatcar-key,disable-afterwards' -- "${@}")
 eval set -- "${opts}"
 
 USER_PAYLOAD=
 PAYLOAD=
 VERSION=
+EXTENSIONS=()
 FORCE_DEV_KEY=
 FORCE_FLATCAR_KEY=
 DISABLE_AFTERWARDS=
@@ -17,20 +18,24 @@ LISTEN_PORT_2=9091
 while true; do
   case "$1" in
   -h|--help)
-    echo "Usage: $(basename "${0}") --to-version VERSION [--to-payload FILENAME] [--listen-port-1 PORT] [--listen-port-2 PORT] [--force-dev-key|--force-flatcar-key|--disable-afterwards]"
+    echo "Usage: $(basename "${0}") --to-version VERSION [--to-payload FILENAME [--extension FILENAME]...] [--listen-port-1 PORT] [--listen-port-2 PORT] [--force-dev-key|--force-flatcar-key|--disable-afterwards]"
     echo "  Updates Flatcar Container Linux through a temporary local update service on localhost."
     echo "  The update-engine service will be unmasked (to disable updates again use -A)."
     echo "  The reboot should be done after applying the update, either manually or through your reboot manager (check locksmithd/FLUO)."
     echo "  An error will be reported if a previously applied update wasn't booted into yet (you may discard it with 'update_engine_client -reset_status')."
     echo "  Warning: If you jump between channels, delete any GROUP configured in /etc/flatcar/update.conf for the new defaults to apply."
     echo "Options:"
-    echo "  -V, --to-version <VERSION>	Updates to the version, by default using the matching release from update.release.flatcar-linux.net"
-    echo "  -P, --to-payload <FILENAME>	Updates to the given update payload file instead of downloading it"
-    echo "  -D, --force-dev-key		Bind-mounts the dev key over /usr/share/update_engine/update-payload-key.pub.pem"
-    echo "  -F, --force-flatcar-key	Bind-mounts the Flatcar release key over /usr/share/update_engine/update-payload-key.pub.pem"
-    echo "  -A, --disable-afterwards	Writes SERVER=disabled to /etc/flatcar/update.conf when done (this overwrites any custom SERVER)"
-    echo "  -L, --listen-port-1 <PORT>  Overwrites standard listen port 9090"
-    echo "  -M, --listen-port-2 <PORT>  Overwrites standard listen port 9091"
+    echo "  -V, --to-version <VERSION>		Updates to the version, by default using the matching release from update.release.flatcar-linux.net"
+    echo "  -P, --to-payload <FILENAME>		Updates to the given Flatcar base update payload file instead of downloading it"
+    echo "					(filename does not matter and internally flatcar_production_update.gz is used)"
+    echo "  -E, --extension <FILENAME>		Provides the given extension image as part of the update, required for -P if the system needs an OEM"
+    echo "					or a Flatcar extension, can/must be specified multiple times (filename matters and should end with"
+    echo "					either oem-OEMID.gz or flatcar-NAME.gz)"
+    echo "  -D, --force-dev-key			Bind-mounts the dev key over /usr/share/update_engine/update-payload-key.pub.pem"
+    echo "  -F, --force-flatcar-key		Bind-mounts the Flatcar release key over /usr/share/update_engine/update-payload-key.pub.pem"
+    echo "  -A, --disable-afterwards		Writes SERVER=disabled to /etc/flatcar/update.conf when done (this overwrites any custom SERVER)"
+    echo "  -L, --listen-port-1 <PORT>		Overwrites standard listen port 9090"
+    echo "  -M, --listen-port-2 <PORT>		Overwrites standard listen port 9091"
     echo
     echo "Example for updating to the latest Stable release and disabling automatic updates afterwards:"
     echo '  VER=$(curl -fsSL https://stable.release.flatcar-linux.net/amd64-usr/current/version.txt | grep FLATCAR_VERSION= | cut -d = -f 2)'
@@ -49,6 +54,16 @@ while true; do
       echo "Error: --to-payload must not have an empty value" > /dev/stderr ; exit 1
     fi
     ;;
+  -E|--extension)
+    shift
+    if [ "$1" = "" ]; then
+      echo "Error: --extension must not have an empty value" > /dev/stderr ; exit 1
+    fi
+    if [[ ! "$(basename -- "$1")" =~ ^(flatcar|oem).*gz$ ]]; then
+      echo "Error: --extension expects paths to files named oem-OEMID.gz or flatcar-NAME.gz (with possible 'flatcar_test_update-' prefix), found: $1" > /dev/stderr ; exit 1
+    fi
+    EXTENSIONS+=("$1")
+    ;;
   -L|--listen-port-1)
     shift
     LISTEN_PORT_1="$1"
@@ -58,7 +73,7 @@ while true; do
     ;;
   -M|--listen-port-2)
     shift
-    LISTEN_PORT_2="$1"    
+    LISTEN_PORT_2="$1"
     if [ "$LISTEN_PORT_2" = "" ]; then
       echo "Error: --listen-port-2 must not have an empty value" > /dev/stderr ; exit 1
     fi
@@ -81,6 +96,14 @@ while true; do
   shift
 done
 
+if [ "$#" != 0 ]; then
+  echo "Error: unexpected extra argumuents: $*" > /dev/stderr ; exit 1
+fi
+
+if [ "$PAYLOAD" = "" ] && [ "${#EXTENSIONS[@]}" != 0 ]; then
+  echo "Error: local extensions are only supported with --to-payload" > /dev/stderr ; exit 1
+fi
+
 if [ "${VERSION}" = "" ]; then
   echo "Error: must specify --to-version" > /dev/stderr ; exit 1
 fi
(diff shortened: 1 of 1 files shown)


bin/flatcar-update
```diff
@@ -17,20 +18,23 @@ LISTEN_PORT_2=9091
...
-    echo "  -D, --force-dev-key		Bind-mounts the dev key over /usr/share/update_engine/update-payload-key.pub.pem"
-    echo "  -F, --force-flatcar-key	Bind-mounts the Flatcar release key over /usr/share/update_engine/update-payload-key.pub.pem"
-    echo "  -A, --disable-afterwards	Writes SERVER=disabled to /etc/flatcar/update.conf when done (this overwrites any custom SERVER)"
-    echo "  -L, --listen-port-1 <PORT>  Overwrites standard listen port 9090"
-    echo "  -M, --listen-port-2 <PORT>  Overwrites standard listen port 9091"
+    echo "  -V, --to-version <VERSION>		Updates to the version, by default using the matching release from update.release.flatcar-linux.net"
+    echo "  -P, --to-payload <FILENAME>		Updates to the given Flatcar base update payload file instead of downloading it"
+    echo "					(filename does not matter and internally flatcar_production_update.gz is used)"
+    echo "  -E, --extensions <FILENAME>...	Provides the given extension images as part of the update, required for -P if the system needs an OEM"
+    echo "					or a Flatcar extension, space separated (filename matters and should be either oem-OEMID.gz or flatcar-NAME.gz)"
```
krnowak: I'd change the `--extensions <FILENAME>...` flag to `--extension <FILENAME>` (no ellipsis). The help message should just say that this flag can be specified multiple times (`-E oem-qemu.gz -E flatcar-foo.gz -E flatcar-bar.gz`). Otherwise the invocation of `flatcar-update` can look weird (`flatcar-update -E flatcar-foo.gz -L 4044 -M 2022 -P payload.gz flatcar-baz.gz flatcar-stuff.gz`).

bin/flatcar-update
```diff
@@ -17,20 +18,23 @@ LISTEN_PORT_2=9091
...
+    echo "  -D, --force-dev-key			Bind-mounts the dev key over /usr/share/update_engine/update-payload-key.pub.pem"
+    echo "  -F, --force-flatcar-key		Bind-mounts the Flatcar release key over /usr/share/update_engine/update-payload-key.pub.pem"
+    echo "  -A, --disable-afterwards		Writes SERVER=disabled to /etc/flatcar/update.conf when done (this overwrites any custom SERVER)"
+    echo "  -L, --listen-port-1 <PORT>  		Overwrites standard listen port 9090"
+    echo "  -M, --listen-port-2 <PORT>  		Overwrites standard listen port 9091"
```
krnowak: Indentation looks off for `-L` and `-M` on github. Maybe because they have two spaces before tabs instead of just tabs.

bin/flatcar-update
```diff
@@ -81,6 +95,22 @@ while true; do
   shift
 done
 
+if [ "${#EXTENSIONS[@]}" != 0 ]; then
+  # Allow space separated passing
+  for EXT in "$@"; do
+    EXTENSIONS+=("$EXT")
+    shift
+  done
+fi
```
krnowak: I'd drop it in favor of requiring multiple `-E` parameters. Also, it's missing the check for `oem-*.gz` or `flatcar-*.gz`.

bin/flatcar-update
```diff
@@ -81,6 +95,22 @@ while true; do
...
+
+if [ "${#EXTENSIONS[@]}" = 0 ] && [ "$#" != 0 ]; then
```
krnowak:
```suggestion
if [ "$#" != 0 ]; then
```

bin/flatcar-update
```diff
@@ -89,6 +119,32 @@ if [ "${FORCE_DEV_KEY}" = "1" ] && [ "${FORCE_FLATCAR_KEY}" = "1" ]; then
   echo "Error: must only specify one of --force-dev-key or --force-flatcar-key" > /dev/stderr ; exit 1
 fi
 
+# Use the old mount point for compatibility with old instances, where the script gets copied to
+OEMID=$({ grep -m 1 -o "^ID=.*" /usr/share/oem/oem-release 2> /dev/null || true ; } | cut -d = -f 2)
```
krnowak, line 116:
```suggestion
OEMID=$({ grep -m 1 -o "^ID=.*" /usr/share/oem/oem-release 2> /dev/null || true ; } | cut -d = -f 2-)
```

bin/flatcar-update
```diff
@@ -89,6 +119,32 @@ if [ "${FORCE_DEV_KEY}" = "1" ] && [ "${FORCE_FLATCAR_KEY}" = "1" ]; then
...
+# and to prevent errors as early as possible when updating an airgapped instance
+declare -A OEM_SYSEXTS
+OEM_SYSEXTS[qemu]=""
+OEM_SYSEXTS[azure]=""
+
+# Determine what to download from release server if no local payload is given
+if [ "${OEMID}" != "" ] && { [ "${OEM_SYSEXTS[${OEMID}]+_}" ] || [ -e "/usr/share/oem/sysext/active-oem-${OEMID}" ]; }; then
+  if [ "$PAYLOAD" = "" ]; then
+    EXTENSIONS+=("/var/tmp/flatcar-update/oem-${OEMID}.gz")
+  elif ! echo " ${EXTENSIONS[*]}" | tr '/' ' ' | grep -q " oem-${OEMID}.gz"; then # Prefixed with space to disallow prefixes like "flatcar_test_update-"
```
krnowak: How about the following, should work as well and it's simpler. Also makes sure that won't accept `oem-foo.gz2`.
```suggestion
  elif ! echo " ${EXTENSIONS[*]} " | grep -q "[ /]oem-${OEMID}.gz "; then # Prefixed with space to disallow prefixes like "flatcar_test_update-"
```

bin/flatcar-update
```diff
@@ -89,6 +119,32 @@ if [ "${FORCE_DEV_KEY}" = "1" ] && [ "${FORCE_FLATCAR_KEY}" = "1" ]; then
...
+    echo "Error: system requires '${OEMID}' OEM extension but not passed in --extensions" > /dev/stderr ; exit 1
```
krnowak:
```suggestion
    echo "Error: system requires '${OEMID}' OEM extension but not passed in --extension" > /dev/stderr ; exit 1
```

bin/flatcar-update
```diff
@@ -49,6 +53,16 @@ while true; do
       echo "Error: --to-payload must not have an empty value" > /dev/stderr ; exit 1
     fi
     ;;
+  -E|--extensions)
+    shift
+    if [ "$1" = "" ]; then
+      echo "Error: --extensions must not have an empty value" > /dev/stderr ; exit 1
```
krnowak:
```suggestion
      echo "Error: --extension must not have an empty value" > /dev/stderr ; exit 1
```

bin/flatcar-update
```diff
@@ -49,6 +53,16 @@ while true; do
       echo "Error: --to-payload must not have an empty value" > /dev/stderr ; exit 1
     fi
     ;;
+  -E|--extensions)
```
krnowak:
```suggestion
  -E|--extension)
```

bin/flatcar-update
```diff
@@ -49,6 +53,16 @@ while true; do
...
+    shift
+    if [ "$1" = "" ]; then
+      echo "Error: --extensions must not have an empty value" > /dev/stderr ; exit 1
+    fi
+    if [[ ! "$(basename -- "$1")" =~ ^(flatcar|oem).*gz$ ]]; then
+      echo "Error: --extensions expects paths to files named oem-OEMID.gz or flatcar-NAME.gz, found: $1" > /dev/stderr ; exit 1
```
krnowak:
```suggestion
      echo "Error: --extension expects paths to files named oem-OEMID.gz or flatcar-NAME.gz, found: $1" > /dev/stderr ; exit 1
```

bin/flatcar-update
```diff
@@ -89,6 +119,32 @@ if [ "${FORCE_DEV_KEY}" = "1" ] && [ "${FORCE_FLATCAR_KEY}" = "1" ]; then
...
 
+# Use the old mount point for compatibility with old instances, where the script gets copied to
+OEMID=$({ grep -m 1 -o "^ID=.*" /usr/share/oem/oem-release 2> /dev/null || true ; } | cut -d = -f 2)
+
+# TODO: Keep in sync with flatcar-postinst from update-engine (but here we only need an empty string as entry)
+# This list is only best-effort and meant to help with updating old instances that aren't (fully) migrated yet
+# and to prevent errors as early as possible when updating an airgapped instance
+declare -A OEM_SYSEXTS
+OEM_SYSEXTS[qemu]=""
+OEM_SYSEXTS[azure]=""
```
krnowak: I don't like it that much. Would there be a way to avoid having multiple places that need to be kept in sync as we convert OEM's to sysexts? Like downloading `flatcar-postinst` and grepping it or something?

bin/flatcar-update
```diff
@@ -89,6 +119,32 @@ if [ "${FORCE_DEV_KEY}" = "1" ] && [ "${FORCE_FLATCAR_KEY}" = "1" ]; then
...
+
+# Determine what to download from release server if no local payload is given
+if [ "${OEMID}" != "" ] && { [ "${OEM_SYSEXTS[${OEMID}]+_}" ] || [ -e "/usr/share/oem/sysext/active-oem-${OEMID}" ]; }; then
+  if [ "$PAYLOAD" = "" ]; then
+    EXTENSIONS+=("/var/tmp/flatcar-update/oem-${OEMID}.gz")
+  elif ! echo " ${EXTENSIONS[*]}" | tr '/' ' ' | grep -q " oem-${OEMID}.gz"; then # Prefixed with space to disallow prefixes like "flatcar_test_update-"
+    echo "Error: system requires '${OEMID}' OEM extension but not passed in --extensions" > /dev/stderr ; exit 1
+  fi
+fi
+for NAME in $(grep -h -o '^[^#]*' /etc/flatcar/enabled-sysext.conf /usr/share/flatcar/enabled-sysext.conf 2> /dev/null | grep -v -x -f <(grep '^-' /etc/flatcar/enabled-sysext.conf 2> /dev/null | cut -d - -f 2-) | grep -v -P '^(-).*'); do
```
krnowak, line 128: You loooooove grep, don't you? :) (`grep | grep <(grep) | grep`).

On a serious note - a comment would be nice here. As I understand it, it gets a list of enabled/disabled sysexts from `/etc` and `/usr`, then filters out disabled sysexts (prefixed with `-`) based on `/etc`, and drops the disabled sysexts again. Would the following work too?
```suggestion
for NAME in $(grep -h -v '^[^#-]' /etc/flatcar/enabled-sysext.conf /usr/share/flatcar/enabled-sysext.conf 2> /dev/null | grep -v -x -f <(grep '^-' /etc/flatcar/enabled-sysext.conf 2> /dev/null | cut -d - -f 2-)); do
```

bin/flatcar-update
```diff
@@ -89,6 +119,32 @@ if [ "${FORCE_DEV_KEY}" = "1" ] && [ "${FORCE_FLATCAR_KEY}" = "1" ]; then
...
+  if [ "$PAYLOAD" = "" ]; then
+    EXTENSIONS+=("/var/tmp/flatcar-update/flatcar-${NAME}.gz")
+  elif ! echo "${EXTENSIONS[*]}" | grep -q "flatcar-${NAME}.gz"; then
```
krnowak:
```suggestion
  elif ! echo " ${EXTENSIONS[*]} " | grep -q " flatcar-${NAME}.gz "; then
```

bin/flatcar-update
```diff
@@ -1,13 +1,14 @@
 #!/bin/bash
 set -euo pipefail
 
-opts=$(getopt --name "$(basename "${0}")" --options 'hV:P:L:M:DFA' \
-       --longoptions 'help,to-version:,to-payload:,listen-port-1:,listen-port-2:,force-dev-key,force-flatcar-key,disable-afterwards' -- "${@}")
+opts=$(getopt --name "$(basename "${0}")" --options 'hV:P:E:L:M:DFA' \
+       --longoptions 'help,to-version:,to-payload:,extensions:,listen-port-1:,listen-port-2:,force-dev-key,force-flatcar-key,disable-afterwards' -- "${@}")
```
krnowak:
```suggestion
       --longoptions 'help,to-version:,to-payload:,extension:,listen-port-1:,listen-port-2:,force-dev-key,force-flatcar-key,disable-afterwards' -- "${@}")
```

bin/flatcar-update
```diff
@@ -17,20 +18,23 @@ LISTEN_PORT_2=9091
 while true; do
   case "$1" in
   -h|--help)
-    echo "Usage: $(basename "${0}") --to-version VERSION [--to-payload FILENAME] [--listen-port-1 PORT] [--listen-port-2 PORT] [--force-dev-key|--force-flatcar-key|--disable-afterwards]"
+    echo "Usage: $(basename "${0}") --to-version VERSION [--to-payload FILENAME [--extensions FILENAME...]] [--listen-port-1 PORT] [--listen-port-2 PORT] [--force-dev-key|--force-flatcar-key|--disable-afterwards]"
```
krnowak:
```suggestion
    echo "Usage: $(basename "${0}") --to-version VERSION [--to-payload FILENAME [--extension FILENAME]] [--listen-port-1 PORT] [--listen-port-2 PORT] [--force-dev-key|--force-flatcar-key|--disable-afterwards]"
```

bin/flatcar-update
```diff
@@ -17,20 +18,23 @@ LISTEN_PORT_2=9091
...
-    echo "  -D, --force-dev-key		Bind-mounts the dev key over /usr/share/update_engine/update-payload-key.pub.pem"
-    echo "  -F, --force-flatcar-key	Bind-mounts the Flatcar release key over /usr/share/update_engine/update-payload-key.pub.pem"
-    echo "  -A, --disable-afterwards	Writes SERVER=disabled to /etc/flatcar/update.conf when done (this overwrites any custom SERVER)"
-    echo "  -L, --listen-port-1 <PORT>  Overwrites standard listen port 9090"
-    echo "  -M, --listen-port-2 <PORT>  Overwrites standard listen port 9091"
+    echo "  -V, --to-version <VERSION>		Updates to the version, by default using the matching release from update.release.flatcar-linux.net"
+    echo "  -P, --to-payload <FILENAME>		Updates to the given Flatcar base update payload file instead of downloading it"
+    echo "					(filename does not matter and internally flatcar_production_update.gz is used)"
+    echo "  -E, --extensions <FILENAME>...	Provides the given extension images as part of the update, required for -P if the system needs an OEM"
+    echo "					or a Flatcar extension, space separated (filename matters and should end with either oem-OEMID.gz or flatcar-NAME.gz)"
```
krnowak: I'd change the `--extensions <FILENAME>...` flag to `--extension <FILENAME>` (no ellipsis). The help message should just say that this flag can be specified multiple times (`-E oem-qemu.gz -E flatcar-foo.gz -E flatcar-bar.gz`). Otherwise the invocation of `flatcar-update` can look weird (`flatcar-update -E flatcar-foo.gz -L 4044 -M 2022 -P payload.gz flatcar-baz.gz flatcar-stuff.gz`).

bin/flatcar-update
```diff
@@ -89,6 +119,32 @@ if [ "${FORCE_DEV_KEY}" = "1" ] && [ "${FORCE_FLATCAR_KEY}" = "1" ]; then
...
+# and to prevent errors as early as possible when updating an airgapped instance
+declare -A OEM_SYSEXTS
+OEM_SYSEXTS[qemu]=""
+OEM_SYSEXTS[azure]=""
+
+# Determine what to download from release server if no local payload is given
+if [ "${OEMID}" != "" ] && { [ "${OEM_SYSEXTS[${OEMID}]+_}" ] || [ -e "/usr/share/oem/sysext/active-oem-${OEMID}" ]; }; then
+  if [ "$PAYLOAD" = "" ]; then
+    EXTENSIONS+=("/var/tmp/flatcar-update/oem-${OEMID}.gz")
+  elif ! echo " ${EXTENSIONS[*]}" | tr '/' ' ' | grep -q -P " (flatcar_test_update-|)oem-${OEMID}.gz"; then # Prefixed with space to only match base name
```
krnowak: How about the following, should work as well and it's simpler. Also makes sure that won't accept `oem-foo.gz2`.
```suggestion
  elif ! echo " ${EXTENSIONS[*]} " | grep -q -P "[ /](flatcar_test_update-)?oem-${OEMID}.gz "; then # Prefixed with space to only match base name
```

bin/flatcar-update
```diff
@@ -89,6 +119,32 @@ if [ "${FORCE_DEV_KEY}" = "1" ] && [ "${FORCE_FLATCAR_KEY}" = "1" ]; then
...
+    echo "Error: system requires '${OEMID}' OEM extension but not passed in --extensions" > /dev/stderr ; exit 1
+  fi
+fi
+for NAME in $(grep -h -o '^[^#]*' /etc/flatcar/enabled-sysext.conf /usr/share/flatcar/enabled-sysext.conf 2> /dev/null | grep -v -x -f <(grep '^-' /etc/flatcar/enabled-sysext.conf 2> /dev/null | cut -d - -f 2-) | grep -v -P '^(-).*'); do
+  if [ "$PAYLOAD" = "" ]; then
+    EXTENSIONS+=("/var/tmp/flatcar-update/flatcar-${NAME}.gz")
+  elif ! echo "${EXTENSIONS[*]}" | tr '/' ' ' | grep -q -P " (flatcar_test_update-|)flatcar-${NAME}.gz"; then
```
krnowak:
```suggestion
  elif ! echo " ${EXTENSIONS[*]} " | grep -q -P "[ /](flatcar_test_update-)?flatcar-${NAME}.gz "; then
```

bin/flatcar-update
krnowak: Ignore this, I posted a refreshed version of this comment.

bin/flatcar-update
krnowak: Ignore this, I posted a refreshed version of this comment.

bin/flatcar-update
pothos: Thanks, yes, I thought about it but wasn't sure if it's better, good to have a second opinion.

bin/flatcar-update
krnowak: My suggestion assumes that comments may happen only in the beginning of the line, not after the extension name. I'd say it's fair as I don't see a point in complicating the file format too much.

bin/flatcar-update
pothos: Thanks!

bin/flatcar-update
pothos: Ok, makes sense, I'll try that - I was kind of surprised to see that multiple args aren't supported and I thought that was an ok workaround^^

bin/flatcar-update
pothos: The delimiter is `=` and not supposed to be part of the OEMID, or?

krnowak: I hope so. :) Another option here is (if you want):
```suggestion
OEMID=$(source /usr/share/oem/oem-release || :; echo "${ID:-}")
```

krnowak (approved):

bin/flatcar-update
pothos: Good idea, since this is used at other places, I think we should change it everywhere in a follow-up

pothos: Tested it and it breaks the main script execution if `oem-release` would have an invalid syntax.

pothos: What works is this here: `$(sh -c "source /usr/share/oem/oem-release" || :; echo "${ID:-}")`

krnowak: > Tested it and it breaks the main script execution if `oem-release` would have an invalid syntax.

Alright, let's have a solution that works. I'll have a look at making sure that this file is valid for sourcing.

pothos: I'm sure it's valid as we ship it but if it's broken due to invalid user changes I wanted the script not to crash, hence the `sh -c` workaround (since it's a subshell, `|| :` won't be needed then).

pothos: Ah, no, my example is broken because the echo should be part of the `sh -c`

pothos: This here works `ID=$(sh -c 'source /usr/share/oem/oem-release; echo "${ID:-}"' 2>/dev/null || true)`

PR #103, opened 2023-09-07
systemd/system: add sshkeys@.service

tormath1:
this service allows to conditionnally start the
`coreos-metadata-sshkeys@.service` on DO, EM, EC2 and GGE with the default `core` instance in order to control the `coreos-metadata-sshkeys@.service` from the generic image and not from the OEM.

---

See: https://github.com/flatcar/scripts/pull/1083 for context, CI and cie.


systemd/system/sshkeys@.service
```diff
@@ -0,0 +1,27 @@
...
+Upholds=coreos-metadata-sshkeys@%i.service
+
+[Service]
+Type=oneshot
+RemainAfterExit=yes
+ExecStart=/bin/true
+
+[Install]
+WantedBy=multi-user.target
+DefaultInstance=core
```
pothos: Do we really need the `@` parameterization? I think that users would directly enable `coreos-metadata-sshkeys@USER.service` for any other users they want to have instead of enabling `sshykeys@USER` to have this start the other one.

pothos: So, would just `sshkeys.service` be enough with `Upholds=coreos-metadata-sshkeys@core.service` hardcoded?

pothos (approved):
One question but even in the current state it's good

pothos:
It still has the `@` in it, I think the file name for service and symlink could be `sshkeys.service`

tormath1:
ah 🤦

PR #104, opened 2023-09-08
systemd/system: use sshkeys.service without parameters

tormath1:
https://github.com/flatcar/init/pull/103#issuecomment-1711418735 follow-up (sorry about that)


pothos (approved):
No problem, it would have still worked if I understand correctly^^
Thanks for the quick follow-up

PR #106, opened 2023-09-12
systemd/network: Ignore mana VFs on Azure

jepio:
# Ignore mana VFs on Azure

Azure is introducing a new generation of NICs, that are architected the same way as the current Mellanox based ones: there is a synthetic (vmbus) NIC for handling some control traffic (including DHCP) and a VF that is enslaved to the synthetic NIC. We therefore need to exclude the devices manage by the mana driver from IP management.

## How to use

## Testing done

No testing, just got a LISA bugreport.

### DIFF
--- a/systemd/network/yy-azure-sriov-coreos.network
+++ b/systemd/network/yy-azure-sriov-coreos.network
@@ -9,7 +9,7 @@ KernelCommandLine=coreos.oem.id=azure
 # VF driver currently used in Azure. If other drivers come into use, the
 # symptom will be a VF interface in the output of "networkctl" which never
 # finishes configuring.
-Driver=mlx4_en mlx5_core
+Driver=mlx4_en mlx5_core mana
 
 [Link]
 Unmanaged=yes
--- a/systemd/network/yy-azure-sriov.network
+++ b/systemd/network/yy-azure-sriov.network
@@ -9,7 +9,7 @@ KernelCommandLine=flatcar.oem.id=azure
 # VF driver currently used in Azure. If other drivers come into use, the
 # symptom will be a VF interface in the output of "networkctl" which never
 # finishes configuring.
-Driver=mlx4_en mlx5_core
+Driver=mlx4_en mlx5_core mana
 
 [Link]
 Unmanaged=yes


dongsupark (approved):

pothos (commented):
Reminds me that this file is something afterburn could generate based on the MAC address instead of having to chase driver changes.

pothos:
> Reminds me that this file is something afterburn could generate based on the MAC address instead of having to chase driver changes.

Existing issue for that: https://github.com/flatcar/Flatcar/issues/554

jepio:
interesting - my initial thought was "i can implement that" but then i realised we also need it to work for hotplugged interfaces, so we'd need to re-run afterburn if we did this.

how about i just configure the list to say `Driver=!hv_netvsc`?

pothos:
> interesting - my initial thought was "i can implement that" but then i realised we also need it to work for hotplugged interfaces, so we'd need to re-run afterburn if we did this.

Ah, the metadata server output will change then? Retriggering afterburn from udev could work in that case.

> how about i just configure the list to say `Driver=!hv_netvsc`?

We still need to restrict the matching to "meant to be physical" NICs. If they are all PCI devices, that can work with matching for the path or something but otherwise we could also take the restriction list from the default rule as this is also for that type of NICs.
So, something like this?
```
# Ignore SR-IOV interface on Azure, since it'll be transparently bonded
# to the synthetic interface
[Match]
KernelCommandLine=flatcar.oem.id=azure
# With NetworkManager, Azure uses a udev rule matching DRIVERS=="hv_pci".
# This won't work with networkd because it only checks the driver of the
# device itself, not its parents. All we can do instead is set the non-hv_netvsc
# NIC to be Unmanaged. In the output of networkctl you should see only
# one virtual NIC configured and the "physical" NIC be bonded to it and set
# Unmanaged.
Name=*
Type=!loopback bridge tunnel vxlan wireguard
Driver=!veth dummy hv_netvsc

[Link]
Unmanaged=yes
```

pothos:
FYI, also needs a PR for bootengine:dracut/03flatcar-network/yy-azure-sriov.network and bootengine:dracut/03flatcar-network/yy-azure-sriov-coreos.network (yeah, two files here and two files there to keep in sync…)

jepio:
Opened the second PR here:
- https://github.com/flatcar/bootengine/pull/70

PR #111, opened 2023-10-12
flatcar-update: Source version.txt to get flatcar version

krnowak:
It guards against possible quotes.


krnowak:
Will get back to it maybe some other time.

PR #112, opened 2023-11-17
systemd/system/sshkeys: Make execution more robust

pothos:
The service failed when the unit to start was masked. Since it involves
    networking, the service can fail and if it does we should restart. It
    also seems that the service relied on
    update-ssh-keys-after-ignition.service to run update-ssh-keys once to
    create the subfolder and could possibly race with it.

Prevent execution when it has no chance to succeed because the unit to
    start is masked but in all other cases try to restart on failure. Also
    order behind update-ssh-keys-after-ignition.service.

## How to use

Backport

## Testing done

with
```
variant: flatcar
version: 1.0.0
systemd:
  units:
    - name: sshkeys.service
      dropins:
        - name: test.conf
          contents: |
            [Unit]
            After=update-ssh-keys-after-ignition.service
            ConditionPathIsSymbolicLink=!/etc/systemd/system/coreos-metadata-sshkeys@core.service
            [Service]
            Restart=on-failure
```

and verifying that the service does not run when the unit to start is masked.
@tormath1 also checked that the spurious failure experienced when `update-ssh-keys-after-ignition.service` is slower than `sshkeys.service` didn't occur again.

### DIFF
--- a/systemd/system/sshkeys.service
+++ b/systemd/system/sshkeys.service
@@ -1,4 +1,9 @@
 [Unit]
+# Don't race and we also want to rely on the folder to be created
+After=update-ssh-keys-after-ignition.service
+
+ConditionPathIsSymbolicLink=!/etc/systemd/system/coreos-metadata-sshkeys@core.service
+
 ConditionKernelCommandLine=|ignition.platform.id=packet
 ConditionKernelCommandLine=|flatcar.oem.id=packet
 ConditionKernelCommandLine=|coreos.oem.id=packet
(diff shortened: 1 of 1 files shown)


systemd/system/sshkeys.service
```diff
@@ -22,6 +22,8 @@ ConditionKernelCommandLine=|coreos.oem.id=openstack
 [Service]
 Type=oneshot
 RemainAfterExit=yes
+Restart=on-failure
+ExecCondition=/usr/bin/bash -c 'if [ $("readlink -f /etc/systemd/system/coreos-metadata-sshkeys@core.service") = "/dev/null" ]; then exit 1; fi'
```
tormath1: I'm wondering if we could not try a condition like:
```
ConditionFileNotEmpty=/etc/systemd/system/coreos-metadata-sshkeys@core.service
or 
ConditionPathIsSymbolicLink=!/etc/systemd/system/coreos-metadata-sshkeys@core.service
```

pothos: Thanks, pushed

tormath1 (approved):

PR #113, opened 2024-01-16
Restrict Permissions of grub/menu.lst

### LINKED ISSUE #296: Permission of grub/menu.lst is not persistent
**Description**

`/boot/boot/grub/menu.lst` permissions are `0755`, and setting it to `0644` or `0600` is gone on reboot.

This could also be fixed by enforcing the permission by whatever is resetting it on boot

**Impact**

Low impact, just that the file is world readable, and doesn't need to be executable

**Environment and steps to reproduce**

1. **Set-up**: tested on 2605.9.0 in qemu
1. **Action(s)**:
  a. `chmod 0600 /boot/boot/grub/menu.lst`
  b. `systemctl reboot`
  c. `stat /boot/boot/grub/menu.lst`

justdan96:
This is described in the following issue:
https://github.com/kinvolk/Flatcar/issues/296

Setting the `Options=umask` parameter as that behaviour is well documented by systemd: https://www.freedesktop.org/software/systemd/man/latest/systemd.mount.html#Options.

# Restrict Permissions of grub/menu.lst

Previously the permissions of grub/menu.lst were 0755, this PR corrects those permissions which helps with CIS compliance.

## How to use

Test new unit file works with Flatcar installation.

## Testing done

None yet.

Closes: https://github.com/flatcar/Flatcar/issues/296

### DIFF
--- a/systemd/system/boot.mount
+++ b/systemd/system/boot.mount
@@ -9,3 +9,4 @@ ConditionPathExists=!/usr/.noupdate
 [Mount]
 What=/dev/disk/by-label/EFI-SYSTEM
 Where=/boot
+Options=umask=0077


chewi:
I've sort of tested this by copying and adding the change to /run/systemd/system/boot.mount. /boot is an autofs, so you can see the result if you don't access it before that. I'll run this through CI though.

chewi:
The `cl.verity` Kola test failed in a relevant way. It's probably a test-specific failure though.

chewi:
[CI has passed](http://jenkins.infra.kinvolk.io:8080/job/container/job/test/36707/cldsv/) with the Kola fix. Just need to get that merged first.

jepio (approved):

tormath1:
Thanks @justdan96 for the proposed fix, and sorry for the delay here - this has been somehow missed. 🙏

PR #116, opened 2024-03-25
udev: Add rule for /dev/ptp_hyperv systemd unit

jepio:
This allows services (such as chrony) to have a dependency on /dev/ptp_hyperv being present.

Inspired by: https://github.com/microsoft/azurelinux/pull/6234.

### DIFF
--- a/udev/rules.d/51-ptp-hyperv.rules
+++ b/udev/rules.d/51-ptp-hyperv.rules
@@ -1,0 +1,1 @@
+ACTION=="add", SUBSYSTEM=="ptp", ATTR{clock_name}=="hyperv", TAG+="systemd"


udev/rules.d/51-ptp-hyperv.rules
```diff
@@ -0,0 +1 @@
+ACTION=="add", SUBSYSTEM=="ptp", ATTR{clock_name}=="hyperv", TAG+="systemd"
```
pothos, line 1: Is there no upstream place where this rule could live?

pothos (approved):

udev/rules.d/51-ptp-hyperv.rules
jepio: unfortunately not

pothos: Wouldn't this be something for the systemd repo?

jepio: i think not - it's specific to our chronyd configuration.

but if this comes up again or we want to upstream this to other distros we'll find a place then.

jepio:
@pothos this rule is actually unnecessary since systemd v255: https://github.com/systemd/systemd/pull/28973/files

PR #118, opened 2024-04-11
afterburn: set ssh keys for hetzner OEM

apricote:
# Use SSH Keys from metadata service for Hetzner OEM

Enables the existing afterburn service for the `hetzner` OEM. This will read the SSH Key(s) that the user configured for the server from the metadata service, and add them to the `core` user.

## How to use

Changes are pulled into `flatcar/scripts` in https://github.com/flatcar/scripts/pull/1880. This also explains how to use/test them.

## Testing done

I have referenced this commit in `flatcar/scripts` overlay and built a test image. A server created from this image did correctly read and configure the SSH Key. I was able to login and the file `/home/core/.ssh/authorized_keys` contained my key.

RE the above checkboxes, I was not sure if I should add a changelog, some other recent PRs for Proxmox and Akamai did not add any changelog entries.

### DIFF
--- a/systemd/system/sshkeys.service
+++ b/systemd/system/sshkeys.service
@@ -24,6 +24,9 @@ ConditionKernelCommandLine=|ignition.platform.id=openstack
 ConditionKernelCommandLine=|flatcar.oem.id=openstack
 ConditionKernelCommandLine=|coreos.oem.id=openstack
 
+ConditionKernelCommandLine=|ignition.platform.id=hetzner
+ConditionKernelCommandLine=|flatcar.oem.id=hetzner
+
 [Service]
 Type=oneshot
 RemainAfterExit=yes


systemd/system/sshkeys.service
```diff
@@ -24,6 +24,10 @@ ConditionKernelCommandLine=|ignition.platform.id=openstack
 ConditionKernelCommandLine=|flatcar.oem.id=openstack
 ConditionKernelCommandLine=|coreos.oem.id=openstack
 
+ConditionKernelCommandLine=|ignition.platform.id=hetzner
+ConditionKernelCommandLine=|flatcar.oem.id=hetzner
+ConditionKernelCommandLine=|coreos.oem.id=hetzner
```
tormath1:
```suggestion
```
Same reason as in https://github.com/flatcar/scripts/pull/1880#discussion_r1561377054

apricote: Removed in f47c7a60ed0692cf017b0668e4eb441b3c2b9d42

tormath1 (approved):

PR #119, opened 2024-04-12
flatcar-update: Work around a bug in bash 4.3

pothos:
The use of an empty array in bash 4.3 throws an error if set -u is used. This got fixed later but for old clients we need to have a workaround.

## How to use

Update from 2905.2.3 with `sudo ./flatcar-update --to-version 3510.3.2`

## Testing done

The above


krnowak (approved):
Yuck, but meh.

:)

PR #124, opened 2024-07-05
flatcar-update: Replace ncat with socat

krnowak:
This replaces ncat with socat, because ncat will be replaced with some openbsd variant, which doesn't have an option for running a script on the established connection. We already are using socat for serving the payload, so this just follows suit.

Testing:

- spin up a qemu vm with Flatcar 4012.0.1
- inside the vm:
  - `wget https://raw.githubusercontent.com/flatcar/init/krnowak/openbsd-netcat/bin/flatcar-update`
  - `wget https://bincache.flatcar-linux.net/images/amd64/4048.0.0+nightly-20240731-2100/flatcar_test_update.gz`
  - `wget https://bincache.flatcar-linux.net/images/amd64/4048.0.0+nightly-20240731-2100/flatcar_test_update-oem-qemu.gz`
  - `sudo ./flatcar-update --to-version 4048.0.0+nightly-20240731-2100 --to-payload flatcar_test_update.gz --extension flatcar_test_update-oem-qemu.gz --force-dev-key`

### DIFF
--- a/bin/flatcar-update
+++ b/bin/flatcar-update
@@ -267,12 +267,36 @@ tee -a /tmp/response > /dev/null <<-EOF
 	</updatecheck><event status="ok"></event></app></response>
 EOF
 
+# Cleanups for local socat servers below
+
 true > /tmp/payload-server-pids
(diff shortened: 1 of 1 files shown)


chewi (approved):
Not sure why you used `tee` instead of `cat`, but your test worked for me. :smiley:

krnowak:
> Not sure why you used `tee` instead of `cat`, but your test worked for me. 😃

Copy pasta from the socat setup for the payload server. :)

PR #125, opened 2024-09-11
flatcar-install: make sure `blkid` returns one device

### LINKED ISSUE #1506: flatcar-install fails when using path disk devices with multiple copies
## Description

flatcar-install fails when using path disk devices with multiple copies

ie:

```
# ls -l /dev/disk/by-path/
total 0
lrwxrwxrwx. 1 root root  9 Jul 18 14:01 pci-0000:44:00.0-ata-1 -> ../../sda
lrwxrwxrwx. 1 root root 10 Jul 18 14:01 pci-0000:44:00.0-ata-1-part1 -> ../../sda1
lrwxrwxrwx. 1 root root 10 Jul 18 14:01 pci-0000:44:00.0-ata-1-part2 -> ../../sda2
lrwxrwxrwx. 1 root root 10 Jul 18 14:01 pci-0000:44:00.0-ata-1-part3 -> ../../sda3
lrwxrwxrwx. 1 root root 10 Jul 18 14:01 pci-0000:44:00.0-ata-1-part4 -> ../../sda4
lrwxrwxrwx. 1 root root  9 Jul 18 14:01 pci-0000:44:00.0-ata-1.0 -> ../../sda
lrwxrwxrwx. 1 root root 10 Jul 18 14:01 pci-0000:44:00.0-ata-1.0-part1 -> ../../sda1
lrwxrwxrwx. 1 root root 10 Jul 18 14:01 pci-0000:44:00.0-ata-1.0-part2 -> ../../sda2
lrwxrwxrwx. 1 root root 10 Jul 18 14:01 pci-0000:44:00.0-ata-1.0-part3 -> ../../sda3
lrwxrwxrwx. 1 root root 10 Jul 18 14:01 pci-0000:44:00.0-ata-1.0-part4 -> ../../sda4
lrwxrwxrwx. 1 root root  9 Jul 18 14:01 pci-0000:44:00.0-ata-2 -> ../../sdb
lrwxrwxrwx. 1 root root  9 Jul 18 14:01 pci-0000:44:00.0-ata-2.0 -> ../../sdb
[...]
```

## Impact

```
flatcar-install -d /dev/disk-by-path/pci-0000:44:00.0-ata-1 -i provider.ign
```

Install script fails to mount the OEM partition to write the ignition file

## Expected behavior

Install script completes successfully.

## Additional information

The following command does work:

```
flatcar-install -d /dev/disk-by-path/pci-0000:44:00.0-ata-1.0 -i provider.ign
```

It seems the root cause is the way the OEM partition is located using `blkid` to write the Ignition file.

```sh
local OEM_DEV=$(blkid -t "LABEL=OEM" -o device "${DEVICE}"*)
```

This command returns multiple values in our case but handled like a single path, which consequently fails.

ader1990:
In some cases where path disk devices have multiple copies, `blkid` should return only one device for label.

According to https://linux.die.net/man/8/blkid, the `-l` would be perfect for this, as it returns the regular block device first.

Fixes: https://github.com/flatcar/Flatcar/issues/1506

### DIFF
--- a/bin/flatcar-install
+++ b/bin/flatcar-install
@@ -716,10 +716,10 @@ function write_cloudinit() {
     (
     # The ROOT partition should be #9 but make no assumptions here!
     # Also don't mount by label directly in case other devices conflict.
-    local ROOT_DEV=$(blkid -t "LABEL=ROOT" -o device "${DEVICE}"*)
+    local ROOT_DEV=$(blkid -l -t "LABEL=ROOT" -o device "${DEVICE}"*)
 
     mkdir -p "${WORKDIR}/rootfs"
-    case $(blkid -t "LABEL=ROOT" -o value -s TYPE "${ROOT_DEV}") in
+    case $(blkid -l -t "LABEL=ROOT" -o value -s TYPE "${ROOT_DEV}") in
       "btrfs") mount -t btrfs -o subvol=root "${ROOT_DEV}" "${WORKDIR}/rootfs" ;;
       *)       mount "${ROOT_DEV}" "${WORKDIR}/rootfs" ;;
     esac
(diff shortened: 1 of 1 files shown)


krnowak:
`man blkid` for `--list-one` says as follows:

> Look up only one device that matches the search parameter specified with the --match-token option. If there are multiple devices that match the specified search parameter, then the device with the highest priority is returned, and/or the first device found at a given priority (but see below note about udev). Device types in order of decreasing priority are: Device Mapper, EVMS, LVM, MD, and finally regular block devices. If this option is not specified, blkid will print all of the devices that match the search parameter.

So looks like regular block device is of lowest priority. I'm not sure if it matters, though - we still want to have just one device, so this seems to be an alright change. What else we could do is to slap ` | head -n1` instead, but this is as justified as your `-l` and arguably hackier. I'm inclined to approve it.

krnowak (approved):

ader1990:
Thanks @krnowak! I will try to look for another alternative this week, and if I do not find a more suitable approach, I will merge it.

PR #126, opened 2024-10-22
OEM: IONOS Cloud Images

tuunit:
# [Title: describe the change in one sentence]

[ describe the change in 1 - 3 paragraphs ]

## How to use

## Testing done

[Describe the testing you have done before submitting this PR. Please include both the commands you issued as well as the output you got.]

### DIFF
--- a/systemd/system/ignition-delete-config.service
+++ b/systemd/system/ignition-delete-config.service
@@ -10,6 +10,7 @@ ConditionKernelCommandLine=|coreos.oem.id=virtualbox
 ConditionKernelCommandLine=|ignition.platform.id=vmware
 ConditionKernelCommandLine=|flatcar.oem.id=vmware
 ConditionKernelCommandLine=|coreos.oem.id=vmware
+ConditionKernelCommandLine=|flatcar.oem.id=ionoscloud
 
 DefaultDependencies=no
 # Run before any user services to prevent potential config leaks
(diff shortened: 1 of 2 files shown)


systemd/system/ignition-delete-config.service
```diff
@@ -10,6 +10,7 @@ ConditionKernelCommandLine=|coreos.oem.id=virtualbox
 ConditionKernelCommandLine=|ignition.platform.id=vmware
 ConditionKernelCommandLine=|flatcar.oem.id=vmware
 ConditionKernelCommandLine=|coreos.oem.id=vmware
+ConditionKernelCommandLine=|flatcar.oem.id=ionoscloud
```
tormath1, line 13: I don't see an implementation for `ignition-rmcfg` in the Ignition PR. 
That said, you could easily implement it as Ignition would just need to remove this file: `/var/lib/cloud/seed/nocloud/user-data` once consumed.

tuunit: but we need to keep it for cloud-init

tormath1: In this case we can drop this line.
```suggestion
```

tuunit: This PR has been closed a couple of days ago 😅 So nothing to merge here anyhow

tuunit: Replaced by https://github.com/flatcar/init/pull/127

tormath1: Ah sorry, I clicked on the link from the `scripts` PR description. :D

PR #128, opened 2025-01-19
Reload systemd-networkd.service during ensure-sysext

jmacdonagh:
# Supports sysexts that configure networkd (for example, tailscale)

Certain sysext extensions, such as tailscale, need to configure networkd (for example, making `tailscale*` interfaces unmanaged).

## Testing done

- Added drop-in that replaced `ensure-sysext` with the above `ExecStart`
- Added newly built `tailscale.raw` from this MR: https://github.com/flatcar/sysext-bakery/pull/106
- Rebooted, checked `systemd-networkd.service` logs to see that it was reconfigured before `tailscale0` interface came up.

### DIFF
--- a/systemd/system/ensure-sysext.service
+++ b/systemd/system/ensure-sysext.service
@@ -12,6 +12,7 @@ ConditionDirectoryNotEmpty=|/usr/lib/extensions
 Type=oneshot
 RemainAfterExit=yes
 ExecStart=/usr/bin/systemctl daemon-reload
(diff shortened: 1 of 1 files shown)


pothos:
Hey, thanks for opening this PR. We want to get rid of these Flatcar-specific workarounds by having vanilla systemd-sysext do more. It can already do the daemon reload which we added for this. For your use case I want to add a field for service restart/reload to the extension-release metadata file. I wonder if a restart of networkd would also work instead of a reload but maybe it's good anyway to support both.

PR #132, opened 2025-06-12
extend-filesystems: Fix race condition by not using lsblk

chewi:
# extend-filesystems: Fix race condition by not using lsblk

lsblk relies on udev, which is inherently racy, especially when the filesystem has just been mounted. It has been observed that the `FSTYPE` field is sometimes not populated, preventing the filesystem from being resized, and causing tests involving the dev container (which is large) to fail.

Use findmnt instead, which gets its information directly from the kernel. It also natively supports filtering by filesystem type and mount option. It doesn't fetch `PARTTYPE`, but we can safely get that from cgpt.

This could be quite an important fix. We have only seen test failures so far, and maybe that's because it only triggers when the system is under load, but I see no reason why it wouldn't happen in production. Perhaps this only started breaking recently following an update to the kernel or udev or something.

## How to use

```
qemu-img create -f qcow2 -o backing_file=flatcar_production_qemu_uefi_image.img,backing_fmt=qcow2,lazy_refcounts=on,size=16106127360 my_flatcar.img
./flatcar_production_qemu_uefi.sh -I my_flatcar.img,snapshot=on 
```

Check that the extend-filesystems.service unit works successfully and that / is actually resized.

## Testing done

[CI has passed](http://jenkins.infra.kinvolk.io:8080/job/container/job/test/36707/cldsv/). I previously forced the test to fail so that it would retry and I could see the output. Once I had applied the fix, it would successfully resize every time, whereas before it would fail most of the time.

### DIFF
--- a/scripts/extend-filesystems
+++ b/scripts/extend-filesystems
@@ -10,56 +10,49 @@
 # the partition but the filesystem resize fails the operation will not
 # be re-attempted by this script later unless the disk grows even more.
 
-set -e -o pipefail
+set -euo pipefail
 
-COREOS_RESIZE="3884dd41-8582-4404-b9a8-e9b84f2df50e"
+COREOS_RESIZE="3884DD41-8582-4404-B9A8-E9B84F2DF50E"
 
-declare -a DEV_LIST
-mapfile DEV_LIST < <(lsblk -P -o NAME,PARTTYPE,FSTYPE,MOUNTPOINT)
-
-for dev_info in "${DEV_LIST[@]}"; do
+while read -r dev_info; do
     eval "$dev_info"
-    NAME=${NAME//\!//}
-    echo "Found ${NAME}"
-    if [[ "${PARTTYPE}" != "${COREOS_RESIZE}" ]] || \
-       [[ -z "${NAME}" ]] || \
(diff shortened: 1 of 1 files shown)


scripts/extend-filesystems
```diff
@@ -10,56 +10,47 @@
...
             ;;
         "xfs")
-            xfs_growfs "${MOUNTPOINT}"
+            xfs_growfs "${SOURCE}"
             ;;
     esac
-    echo "Resized filesystem in ${device}"
-done
+    echo "Resized filesystem in ${SOURCE}"
+done < <(findmnt --types btrfs,ext4,xfs --options rw --pairs --shell --output SOURCE,FSTYPE,TARGET | sort -u -k1,1)
```
jepio, line 58: Why do we need `sort -u -k1,1` (what does it defend against)? If it is essential, could you capture the reason as a comment.

jepio (approved):

scripts/extend-filesystems
chewi: It's so that things like bind mounts don't make it try to resize the same partition more than once. It wouldn't do any real harm, but it seems sensible to avoid it in the first place. I'll add a comment.

PR #134, opened 2025-08-18
fix failing install on recycled LVM disk

cbricart:
`flatcar-install` fails on disks, that previously
have had a valid LVM setup on it.
A valid LVM volume group is auto-activated, locks
the original partition table and hence prevents
it from being re-read after writing the Flatcar
raw disk image to the device.

This patch unconditionally deactivates any possibly present auto-activated LVM volume groups to release the device-mapper claim on the partitions table.

# [Title: describe the change in one sentence]

[ describe the change in 1 - 3 paragraphs ]

## How to use

## Testing done

[Describe the testing you have done before submitting this PR. Please include both the commands you issued as well as the output you got.]

### DIFF
--- a/bin/flatcar-install
+++ b/bin/flatcar-install
@@ -17,7 +17,7 @@ else
     ensure_tool "${BZIP_UTIL}"
 fi
 
-toolset=( blockdev btrfstune cp cut dd gawk gpg grep head ls lsblk mkdir mkfifo mktemp mount rm sed sort tee udevadm wget wipefs ) 
+toolset=( blockdev btrfstune cp cut dd gawk gpg grep head ls lsblk lvm mkdir mkfifo mktemp mount rm sed sort tee udevadm wget wipefs ) 
 
 for t in "${toolset[@]}"; do ensure_tool "${t}"; done
 
@@ -618,6 +618,10 @@ function get_disk_status() {
 
 function write_to_disk() {
     mkfifo -m 0600 "${WORKDIR}/disk_modified"
+
+    # unconditionally try to deactive any possible auto-activated LVM volume groups
+    # as those will block the modified partition table from being re-read.
+    lvm vgchange -an || true
 
     # We are at the point of no return, so wipe disk labels missed below.
     # In particular, ZFS writes labels in the last half-MiB of the disk.


cbricart:
PR as discussed with @t-lo and @tormath1 live at FrOSCon 2025

tormath1:
@flatcar/flatcar-maintainers any chance to have a look? This looks ok to me but I'd like to have a double check.

chewi:
It makes sense to me. Would CI cover this?

tormath1:
> It makes sense to me. Would CI cover this?

In a way, our Equinix Metal tests are using `flatcar-install` so it should catch a regression.

bin/flatcar-install
```diff
@@ -619,6 +619,10 @@ function get_disk_status() {
 function write_to_disk() {
     mkfifo -m 0600 "${WORKDIR}/disk_modified"
 
+    # unconditionally try to deactive any possible auto-activated LVM volume groups
+    # as those will block the modified partition table from being re-read.
+    lvm vgchange -an || true
```
pothos, line 624: Is `lvm vgchange -an` the same as `lvchange -an /dev/mapper/*`?

This is what we had in https://github.com/kinvolk/racker/blob/8faa2de53f3e291db15a372a6c1d3465f18c7ea1/management-node-ipxe/install-ignition.yaml#L28 and I think it's good to include here, thanks for the PR!

cbricart: > Is `lvm vgchange -an` the same as `lvchange -an /dev/mapper/*`?

well, implicitly yes - `lvchange -an …` obviously deactivates any logical volume, while `vgchange -an` works on the volume group level to deactivate, and hence referentially also deactivates any logical volume that would be active on it.

(and calling it via `lvm vgchange …` does not rely on the symlink to the multi-call `lvm` binary being present)

pothos:
> In a way, our Equinix Metal tests are using `flatcar-install` so it should catch a regression.

In the future I guess a new qemu test setup is needed to cover the PXE installation…

pothos (approved):
Thank you

tormath1:
This will be available in next Alpha / Beta. Thanks @cbricart!

PR #135, opened 2025-09-16
sshkeys: enable SSH keys injection from Scaleway IMDS

tormath1:
In this PR, we enable SSH keys injection from Scaleway IMDS. In case one wants to add SSH keys to the instance without using Ignition.

---

I've tested it and this has been tested by the user reporting the issue as well. 

This has to be backported on some maintainance branches for `init`.

### DIFF
--- a/systemd/system/sshkeys.service
+++ b/systemd/system/sshkeys.service
@@ -33,6 +33,9 @@ ConditionKernelCommandLine=|flatcar.oem.id=akamai
 ConditionKernelCommandLine=|ignition.platform.id=proxmoxve
 ConditionKernelCommandLine=|flatcar.oem.id=proxmoxve
 
+ConditionKernelCommandLine=|ignition.platform.id=scaleway
+ConditionKernelCommandLine=|flatcar.oem.id=scaleway
+
 [Service]
 Type=oneshot
 RemainAfterExit=yes


systemd/system/sshkeys.service
```diff
@@ -33,6 +33,8 @@ ConditionKernelCommandLine=|flatcar.oem.id=akamai
 ConditionKernelCommandLine=|ignition.platform.id=proxmoxve
 ConditionKernelCommandLine=|flatcar.oem.id=proxmoxve
 
+ConditionKernelCommandLine=|flatcar.oem.id=scaleway
```
krnowak, line 37: It certainly is an odd one out here, without the ignition.platform.id variant. I suppose it is on purpose, right?

krnowak (approved):

systemd/system/sshkeys.service
tormath1: We initially added this `ignition.platform.id` to be closer from the upstream but I honestly doubt we will change `flatcar.oem.id` -> `ignition.platform.id` one day.
But I'll add the `ignition.platform.id` to be consistent with the other entries (and we never know!)

tormath1:
Cherry-picked to `flatcar-4230-backport`

PR #140, opened 2025-12-29
Drop GCE disk rules and drop Rackspace + EM support

chewi:
The GCE disk rules are now handled by the app-admin/google-guest-configs package.

This also drops support for Rackspace and Equinix Metal (Packet) as we have done elsewhere.

This needs to be merged as part of flatcar/scripts#3606.

## How to use

Nothing to use, this just drops the old rules.

## Testing done

The new rules have already been tested manually. I have run [Jenkins](https://jenkins.flatcar.org/job/container/job/packages_all_arches/38/cldsv/) where GCE failed on some K8s tests, but that's an unrelated issue that I'm fixing elsewhere.


chewi:
Depends on flatcar/scripts#3606.

krnowak (approved):

PR #146, opened 2026-09-04
systemd: Escape %s in oem-cloudinit.service

chewi:
Copilot made a bad suggestion in #145. I had previously tested this with `echo` rather than `printf`. I haven't tested this change, but it can't make things any worse.


tormath1 (approved):
Copilot is curiously silent on this one.

PR #147, opened 2026-09-21
bin/flatcar-install: fix quote issue

tormath1:
We updated /etc/os-release to a more standard way (values are double-quoted) - let's reflect this change here as well.

---

I used the same way as we are doing for Mantle : https://github.com/flatcar/mantle/pull/782

Related to: https://github.com/flatcar/Flatcar/issues/2425


bin/flatcar-install
```diff
@@ -61,8 +61,8 @@ find_smallest_unused_disk() {
 # Everything we do should be user-access only!
 umask 077
 
-if grep -q "^ID=flatcar$" /etc/os-release; then
-    source /etc/os-release
+. /etc/os-release
+if [[ "${ID}" = "flatcar" ]]; then
```
chewi: That isn't doing quite the same thing. I'm not sure why, but it seems like it only wanted to source the file if it is Flatcar. Maybe do something like...

```suggestion
if (. /etc/os-release; [[ ${ID} = flatcar ]]); then
    . /etc/os-release
```

chewi (changes requested):

bin/flatcar-install
tormath1: Yeah, we can do that - I don't know why it wants to source the file only if it's Flatcar, I wanted to simplify a bit the logic here.

chewi (approved):

